PR #33, opened 2017-09-08
additions to support finding users by subject ids while filtering last login by clientid`

kpaul:
tests coming, this is all the supporting code

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/33)

### DIFF
--- a/Fabric.Identity.API/Authorization/ReadAuthorizationHandler.cs
+++ b/Fabric.Identity.API/Authorization/ReadAuthorizationHandler.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Fabric.Identity.API.Configuration;
+using IdentityModel;
+using Microsoft.AspNetCore.Authorization;
+using Serilog;
+
+namespace Fabric.Identity.API.Authorization
+{
+    public class ReadAuthorizationHandler : AuthorizationHandler<ReadScopeRequirement>
+    {
+        private readonly ILogger _logger;
+        private readonly IAppConfiguration _appConfiguration;
+
+        public ReadAuthorizationHandler(IAppConfiguration appConfiguration, ILogger logger)
+        {
+            _logger = logger;
+            _appConfiguration = appConfiguration ?? throw new ArgumentNullException(nameof(appConfiguration)); ;
+        }
+
+        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, ReadScopeRequirement requirement)
+        {
+            //ensure the logged in client has the read scope claim
+
+            if (HasRequiredScopeClaim(context.User, requirement.ReadScope))
+            {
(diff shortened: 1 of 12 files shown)


michaelvidal24:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[Fabric.Identity.API/Management/UsersController.cs, line 29 at r1](https://reviewable.io:443/reviews/healthcatalyst/fabric.identity/33#-Ktl5Thiye1XrPN_F_oX:-Ktl3jDKEkAV4wSsqe2Z:bxbhm74) ([raw file](https://github.com/healthcatalyst/fabric.identity/blob/170c4f067c6acb1c963aa51144c15f1b7d233ecd/Fabric.Identity.API/Management/UsersController.cs#L29)):*
> ```Smalltalk
> 
>         
>         public async Task<IActionResult> Get(string clientId, IEnumerable<string> subjectIds)
> ```

Do we think we have to restrict this method and only allow the client applications to get their own last login date data? Also, I think we'll need to make sure to authorize against a different scope than we do with the other management api's. 

---

*[Fabric.Identity.API/Services/CouchDbAccessService.cs, line 94 at r1](https://reviewable.io:443/reviews/healthcatalyst/fabric.identity/33#-Ktl4v7dSq8MtHq01qX0:-Ktl4v7dSq8MtHq01qX1:bhewp4h) ([raw file](https://github.com/healthcatalyst/fabric.identity/blob/170c4f067c6acb1c963aa51144c15f1b7d233ecd/Fabric.Identity.API/Services/CouchDbAccessService.cs#L94)):*
> ```Smalltalk
>                 {
>                     _logger.Debug($"there was an error getting documents - message: {result.Reason}");
>                     
> ```

Should we throw an exception here? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/healthcatalyst/fabric.identity/33)*

kpaul:
Good call on the exception, I will add that. 

We did talk about adding a new scope for reading user information. I am planning to add that. 

Could you elaborate more on this? 

> Do we think we have to restrict this method and only allow the client applications to get their own last login date data

Are you suggesting we do not accept a client id parameter and instead use the client id of the authenticated client?

michaelvidal24:
> Are you suggesting we do not accept a client id parameter and instead use the client id of the authenticated client?

Yes - I think we should think about it. I can see two use cases:

1. An application needs to get the users and last login activity related to their app
2. A generic auditing/user management app needs to get login activity for all client applications

In the first case, I think we can have use the read scope we were talking about last week, in conjunction with pulling the client id from the access token. In the second case, I think we can use the manageresources scope or a new 'readallclient' scope and have an enpoint that takes in a list of client ids or something along those lines. We don't know those exact requirements yet though, so I wouldn't want to build that until we know what we needed.

michaelvidal24:
After discussing it further, I think it makes sense to leave the function as is since there will be a facade in front of this api call that won't allow us to directly pass the client id via the access token.

PR #34, opened 2017-09-13
User/kpaul/users api tests

kpaul:
added ability to run integration tests against couchdb.
 added integration tests for users api

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/34)


Fabric.Identity.API/Management/UsersController.cs
```diff
@@ -36,7 +43,7 @@ public UsersController(IDocumentDbService documentDbService, AbstractValidator<U
                     HttpStatusCode.NotFound);
```
michaelvidal24: Should we use the CreateFailureResponse here so we can  tell consumers of the API that we are returning a 404 because the client they requested did not exist? As it is now, it will be hard for them to know if it's the client id or the document ids that caused the 404.

kpaul: I cant see it in the view here but this is the client Id check 

`if (string.IsNullOrEmpty(client?.ClientId))
            {
                return CreateFailureResponse($"The specified client with id: {clientId} was not found",
                    HttpStatusCode.NotFound);
            }`

PR #38, opened 2017-09-26
Updated installation script to move parameters to configuration.

michaelvidal24:
This will allow us to simply edit the config parameters in a file, and call the installation script without passing parameters. We also allow pushing a new variable to the config (with an option to encrypt) so it can be used downstream.

This pull request also encompasses moving prerequisite tests and registration steps to the identity install script.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/38)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -1,6 +1,12 @@
 #
 # Install_Identity_Windows.ps1
 #	
+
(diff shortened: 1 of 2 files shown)


michaelvidal24:
The tests passed on my local before making the pull request - these files shouldn't affect those tests anyway.

kpaul:
@michaelvidal24 Those failing tests seem to be related to something I am working on. They passed for me in the build last night so I am not sure why they are failing now but I can look into it.

PR #39, opened 2017-09-26
updated couchdb service to replace backslash with two colons for use in the document id

kpaul:
another minor change was the parameter name for the GetDocuments method

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/39)

### DIFF
--- a/Fabric.Identity.API/Services/CouchDbAccessService.cs
+++ b/Fabric.Identity.API/Services/CouchDbAccessService.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Fabric.Identity.API.Configuration;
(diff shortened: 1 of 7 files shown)


Fabric.Identity.API/Services/CouchDbAccessService.cs
```diff
@@ -22,7 +23,12 @@ public class CouchDbAccessService : IDocumentDbService
 
         private string GetFullDocumentId<T>(string documentId)
```
bsmithvt: Unrelated to your changes, but in the method "Task<T> GetDocument<T>(string documentId)", there are 2 calls to "GetFullDocumentId<T>(documentId)". Perhaps we can consolidate that into a local variable instead of calling it twice.

Fabric.Identity.API/Services/CouchDbAccessService.cs
```diff
@@ -22,7 +23,12 @@ public class CouchDbAccessService : IDocumentDbService
 
         private string GetFullDocumentId<T>(string documentId)
         {
-            return $"{typeof(T).Name.ToLowerInvariant()}:{documentId}";
+            return ReplaceInvalidChars($"{typeof(T).Name.ToLowerInvariant()}:{documentId}");
```
bsmithvt: Should this method just be called on the "documentId" portion of the key, or do we want to include the Type Name as well?

Fabric.Identity.API/Services/CouchDbAccessService.cs
kpaul: good call. will make that change

Fabric.Identity.API/Services/CouchDbAccessService.cs
kpaul: Yeah i see your point. While it doesn't hurt anything to do it on the full string including the type, the type will never contain an invalid character so there will be no replacement.

PR #48, opened 2017-10-06
Added endpoint for getting configured identity providers.

michaelvidal24:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/48)

### DIFF
--- a/Fabric.Identity.API/Quickstart/Account/AccountController.cs
+++ b/Fabric.Identity.API/Quickstart/Account/AccountController.cs
@@ -53,6 +53,7 @@ namespace IdentityServer4.Quickstart.UI
             DocumentDbUserStore documentDbUserStore,
             ILogger logger,
             IExternalIdentityProviderServiceResolver externalIdentityProviderServiceResolver,
+            AccountService account,
             TestUserStore users = null)
         {
             // if the TestUserStore is not in DI, then we'll just use the global users collection
@@ -61,7 +62,7 @@ namespace IdentityServer4.Quickstart.UI
             _events = events;
             _appConfiguration = appConfiguration;
             _logger = logger;
-            _account = new AccountService(interaction, httpContextAccessor, clientStore, appConfiguration);
+            _account = account;
             _userLoginManager = new UserLoginManager(documentDbUserStore, _logger);
             _externalIdentityProviderServiceResolver = externalIdentityProviderServiceResolver;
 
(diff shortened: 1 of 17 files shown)


Fabric.Identity.API/Quickstart/Account/AccountController.cs
```diff
@@ -53,6 +53,7 @@ public class AccountController : Controller
             DocumentDbUserStore documentDbUserStore,
             ILogger logger,
             IExternalIdentityProviderServiceResolver externalIdentityProviderServiceResolver,
+            AccountService account,
```
bsmithvt: Minor - can we name this `accountService` for clarity?

Fabric.Identity.API/Services/IdentityProviderConfigurationService.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _httpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+        }
+
+        public ICollection<ExternalProvider> GetConfiguredIdentityProviders()
+        {
+            var schemes = _httpContextAccessor.HttpContext.Authentication.GetAuthenticationSchemes();
+
+            var providers = schemes
+                .Where(x => x.DisplayName != null && !AccountOptions.WindowsAuthenticationSchemes.Contains(x.AuthenticationScheme))
```
bsmithvt: Should there be a filter on whitespace/empty DisplayName?

Fabric.Identity.IntegrationTests/IdentityProvidersControllerTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Fabric.Identity.IntegrationTests
+{
+    public class IdentityProvidersControllerTests : IntegrationTestsFixture
+    {
+        private readonly string _identityProvidersBaseUrl = "/api/identityProviders";
+        [Fact]
+        public async Task UsersController_Search_InvalidIdentityProvider_ReturnsBadRequest()
```
bsmithvt: Is the test method name accurate? It doesn't appear to match the actual test.

Fabric.Identity.API/Quickstart/Account/AccountController.cs
michaelvidal24: good point, i'll make that change.

Fabric.Identity.API/Services/IdentityProviderConfigurationService.cs
michaelvidal24: In this case, I don't think that is necessary. We are only trying to filter out authentication schemes like Bearer which will have a null display name.

Fabric.Identity.IntegrationTests/IdentityProvidersControllerTests.cs
michaelvidal24: Good catch - i'll fix that.

PR #49, opened 2017-10-07
User/bsmithvt/group filtering

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/49)


Fabric.Identity.API/Startup.cs
```diff
@@ -77,6 +77,13 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton<ILdapConnectionProvider, LdapConnectionProvider>();
             services.AddSingleton<IExternalIdentityProviderServiceResolver, ExternalIdentityProviderServiceResolver>();
             services.AddSingleton<LdapProviderService>();
+
+            if (_appConfig.FilterSettings?.GroupFilterSettings != null)
```
michaelvidal24: Is it necessary to have this check? Can't we just add the GroupFilterSettings, or even the FilterSettings and simply make sure the class checks for null before using it? I think it might be better to keep our registration code as simple as possible.

bsmithvt: @michaelvidal24, I tried that initially, but I get the following error on startup if there are no filters defined:

`System.ArgumentNullException: 'Value cannot be null.'`

PR #56, opened 2017-10-12
added missing IS4 Client model fields to internal API Client model

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/56)


bsmithvt:
@michaelvidal24, that's a good question. If we do that, will we still store the IS4 Client?

michaelvidal24:
We would still want to store the IS4 Client model, as that is the 'domain' model. We would perform a conversion between the API model and the domain model in the API layer. This will allow us to iterate on the domain model and api model independently, providing us the ability to keep changes in the domain model from potentially breaking our API contract.

PR #72, opened 2017-10-30
update response status codes 

kpaul:
when a duplicate is added return a 409
when deleting a non existent resource return 404

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/72)

### DIFF
--- a/Fabric.Identity.API/Management/ClientController.cs
+++ b/Fabric.Identity.API/Management/ClientController.cs
@@ -104,23 +104,23 @@ namespace Fabric.Identity.API.Management
             try
             {
                 var is4Client = client.ToIs4ClientModel();
-                return ValidateAndExecute(is4Client, () =>
-                {
-                    var id = is4Client.ClientId;
+                return ValidateAndExecute(is4Client,
+                    () =>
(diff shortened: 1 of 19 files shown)


kpaul:
Hmm the next run after this failed build passed without any code change. I believe it was some kind of transient failure. I am not sure how to get this to run again without a new commit.

Fabric.Identity.API/Management/ClientController.cs
```diff
@@ -170,6 +170,14 @@ public IActionResult Put(string id, [FromBody] Client client)
         [SwaggerResponse(404, typeof(Error), NotFoundErrorMsg)]
         public IActionResult Delete(string id)
         {
+            var client = _documentDbService.GetDocument<IS4.Client>(id).Result;
+
+            if (string.IsNullOrEmpty(client?.ClientId))
```
michaelvidal24: Seems weird to check for s string to be null or empty when we really want to know if the object is null...I know null propagation will have the result in essentially the same thing because we require ClientId, it just reads strangely to me.

PR #79, opened 2017-11-02
User/bsmithvt/create api models

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/79)

### DIFF
--- a/Fabric.Identity.API/Management/ApiResourceController.cs
+++ b/Fabric.Identity.API/Management/ApiResourceController.cs
@@ -95,24 +95,22 @@ namespace Fabric.Identity.API.Management
         /// <summary>
         ///     Creates an API resource.
         /// </summary>
-        /// <param name="resource">The <see cref="ApiResource" /> object to add.</param>
+        /// <param name="apiResource">The <see cref="ApiResource" /> object to add.</param>
         /// <returns></returns>
         [HttpPost]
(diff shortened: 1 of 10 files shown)


Fabric.Identity.API/Management/ApiResourceController.cs
```diff
@@ -69,95 +69,92 @@ public IActionResult Get(string id)
...
                 {
                     return CreateFailureResponse($"The specified api resource with id:{id} was not found",
                         HttpStatusCode.NotFound);
                 }
 
                 // Prevent from changing secrets.
-                apiResource.ApiSecrets = storedApiResource.ApiSecrets;
+                is4ApiResource.ApiSecrets = storedApiResource.ApiSecrets;
                 // Prevent from changing payload Name.
                 apiResource.Name = id;
```
michaelvidal24: I think this has to be changed to reference is4ApiResource.Name

Fabric.Identity.API/Models/IdentityResource.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace Fabric.Identity.API.Models
+{
+    public class IdentityResource
```
michaelvidal24: Do we need to review the IS4 model defaults here? I think we should use the same defaults.

PR #80, opened 2017-11-02
added sql server client store 

kpaul:
before i continue with the sql server resource stores i wanted to get my approach reviewed.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/80)


Fabric.Identity.API/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -75,7 +75,6 @@ public static void AddFluentValidations(this IServiceCollection serviceCollectio
             serviceCollection.AddTransient<IApiResourceStore, InMemoryApiResourceStore>();
             serviceCollection.AddTransient<IIdentityResourceStore, InMemoryIdentityResourceStore>();
             serviceCollection.AddTransient<IClientManagementStore, InMemoryClientManagementStore>();
-            serviceCollection.AddTransient<IUserStore, CouchDbUserStore>();
```
michaelvidal24: Don't remove this yet, I will fix this up when I fixup the DI setups.

Fabric.Identity.API/Models/Domain/ClientDomainModel.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace Fabric.Identity.API.Models.Domain
+{
+    public class ClientDomainModel : IdentityServer4.EntityFramework.Entities.Client, ISoftDelete, ITrackable
```
michaelvidal24: These models should be in the SqlServer namespace because they are specific to that implementation. the IS4.Models are the domain models.

Fabric.Identity.API/Models/Domain/DomainModelExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using IdentityServer4.EntityFramework.Mappers;
+
+namespace Fabric.Identity.API.Models.Domain
+{
+    public static class DomainModelExtensions
```
michaelvidal24: These should go in the Sql Server namespace too, same with the ITrackable and ISoftDelete interfaces since they are arguably store implementation specific.

Fabric.Identity.API/Models/Domain/ClientDomainModel.cs
kpaul: good call. i will move them

Fabric.Identity.API/Models/Domain/DomainModelExtensions.cs
kpaul: i will move them

PR #83, opened 2017-11-03
add sql server resource stores

kpaul:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/83)


Fabric.Identity.API/Persistence/SqlServer/Models/ApiResourceDomainModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Fabric.Identity.API.Persistence.SqlServer.Models
+{
+    public class ApiResourceDomainModel : IdentityServer4.EntityFramework.Entities.ApiResource, ITrackable, ISoftDelete
```
michaelvidal24: I think we might be better off calling this ApiResourceEntityModel, since it isn't a domain model.

Fabric.Identity.API/Persistence/SqlServer/Models/DomainModelExtensions.cs
```diff
@@ -7,8 +7,19 @@ public static class DomainModelExtensions
         public static ClientDomainModel ToDomainModel(this IdentityServer4.Models.Client is4Client)
```
michaelvidal24: I think we should call these extension methods ToEntityModel

Fabric.Identity.API/Persistence/SqlServer/Models/IdentityResourceDomainModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Fabric.Identity.API.Persistence.SqlServer.Models
+{
+    public class IdentityResourceDomainModel : IdentityServer4.EntityFramework.Entities.IdentityResource, ITrackable, ISoftDelete
```
michaelvidal24: Same comment as above, I think we should call this IdentityResourceEntityModel.

Fabric.Identity.API/Persistence/SqlServer/Models/ApiResourceDomainModel.cs
kpaul: Yeah I thought the same thing and have started changing them over to be called entities

PR #85, opened 2017-11-03
User/bsmithvt/refactor health check

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/85)

### DIFF
--- a/Fabric.Identity.API/Startup.cs
+++ b/Fabric.Identity.API/Startup.cs
@@ -193,7 +193,7 @@ namespace Fabric.Identity.API
             app.UseIdentityServerAuthentication(options);
             app.UseMvcWithDefaultRoute();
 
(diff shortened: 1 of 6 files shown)


Fabric.Identity.API/Startup.cs
```diff
@@ -196,8 +192,10 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
             var options = app.ApplicationServices.GetService<IdentityServerAuthenticationOptions>();
             app.UseIdentityServerAuthentication(options);
             app.UseMvcWithDefaultRoute();
+
+            var healthCheckService = app.ApplicationServices.GetService<IHealthCheckerService>();
```
michaelvidal24: Maybe change this to GetRequiredService so it throws an exception if IHealthCheckerService is not registered? That way we fail faster - on startup rather than when /_monitor/deep is called.

PR #86, opened 2017-11-03
the branch name is misleading as this branch is mostly about the user model 

kpaul:
i started on the user data model. 
framed in the user store but not fully implemented. 
finished up rename of domain model classes to be entity classes which is the real reason to check this in now, so that brian can help with the persisted grant store and not have a nasty merge

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/86)


kpaul:
just realized i forgot to add the new DbSets to the IIDentityDbContent interface. That is changed in the current branch i am working on

PR #88, opened 2017-11-04
added some basic scripts to run tests more easily.

michaelvidal24:
These scripts are designed to be run from the scripts directory. They basically mimic what the CI build and release pipeline do. 

You can run each script individually, or run the run-all-tests.sh script to run unit, integration and functional tests in one shot.

All the scripts except for the run-unit-tests.sh script take two parameters, one for the couchdb username and one for the couchdb password.

You'll need to have docker running, as we spin up couchdb and openldap for the integration tests as a container. For the functional tests we run couchdb, authorization and identity as containers. We build the identity container from the code in your dev environment.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/88)

### DIFF
--- a/Fabric.Identity.API/scripts/run-all-tests.sh
+++ b/Fabric.Identity.API/scripts/run-all-tests.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+couchusername=$1
+couchpassword=$2
+
+./run-unit-tests.sh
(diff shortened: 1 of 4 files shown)


michaelvidal24:
It shouldn't need to be modified, since we will likely use integrated authentication, and we will provide a default server of localhost in the config.

PR #91, opened 2017-11-06
sql server user store implementation

kpaul:
still might need to adjust the entities a bit for FK's but the store implementation is done

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/91)

### DIFF
--- a/Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
+++ b/Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
@@ -18,7 +18,7 @@ namespace Fabric.Identity.API.Persistence.CouchDb.Stores
             _logger = logger;
         }
 
-        public async Task<User> FindBySubjectId(string subjectId)
+        public async Task<User> FindBySubjectIdAsync(string subjectId)
         {
             _logger.Debug($"finding user with subject id: {subjectId}");
             var user = await _documentDbService.GetDocuments<User>(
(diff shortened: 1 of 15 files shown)


Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
```diff
@@ -35,12 +35,12 @@ public CouchDbUserStore(IDocumentDbService documentDbService, ILogger logger)
             return user?.FirstOrDefault();
         }
 
-        public Task<IEnumerable<User>> GetUsersBySubjectId(IEnumerable<string> subjectIds)
+        public Task<IEnumerable<User>> GetUsersBySubjectIdAsync(IEnumerable<string> subjectIds)
```
michaelvidal24: Any reason these are not marked async, and we aren't using the await keyword?

michaelvidal24: Any reason this is not marked async, and we are not awaiting in the method? Is the _documentDbService.GetDocumentsById function not async?

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
```diff
@@ -35,12 +35,12 @@ public CouchDbUserStore(IDocumentDbService documentDbService, ILogger logger)
...
         {
             return _documentDbService.GetDocumentsById<User>(subjectIds);
         }
 
-        public Task<User> AddUser(User user)
+        public Task<User> AddUserAsync(User user)
```
michaelvidal24: Same comment as above.

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
kpaul: i just updated the method name and did not check this file to ensure it was marked async.

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbUserStore.cs
kpaul: same comment as above.

PR #98, opened 2017-11-14
Added a bash script to queue a build from within another build, check…

michaelvidal24:
… its status and download the build artifacts when successfully completed.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/98)

### DIFF
--- a/Fabric.Identity.API/scripts/queue-build.sh
+++ b/Fabric.Identity.API/scripts/queue-build.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+definitionId=$1
+queueBuildResponse=$(curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $SYSTEM_ACCESSTOKEN" -d "{ \"definition\": { \"id\": $definitionId } }" https://healthcatalyst.visualstudio.com/DefaultCollection/Fabric/_apis/build/builds?api-version=2.0)
+echo $queueBuildResponse
+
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/scripts/queue-build.sh
```diff
@@ -0,0 +1,38 @@
...
+definitionId=$1
+queueBuildResponse=$(curl -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $SYSTEM_ACCESSTOKEN" -d "{ \"definition\": { \"id\": $definitionId } }" https://healthcatalyst.visualstudio.com/DefaultCollection/Fabric/_apis/build/builds?api-version=2.0)
+echo $queueBuildResponse
+
+buildId=$(echo ${queueBuildResponse} | jq .id)
+echo "BuildId: $buildId"
+
+status="\"notStarted\""
+completedStatus="\"completed\""
+while [ "$status" != "$completedStatus" ]; do
```
bsmithvt: Are we always guaranteed a "completed" status eventually? Or do we want to put a limit on the number / time elapsed in this loop and fail the parent build once that limit is reached?

michaelvidal24: We'll have a timeout setting on the build task to terminate this step and fail the build if we for some reason don't get a "completed" status.

PR #104, opened 2017-11-27
added code to install the .net core windows server hosting bundle if not found

kpaul:
should we handle the situation where a different version is already installed?

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/104)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -102,10 +102,18 @@ if((Test-Path $zipPackage))
 }
 
 
-if(!(Test-Prerequisite '*.NET Core*Windows Server Hosting*' 1.1.30327.81))
+if(!(Test-Prerequisite-Exact '*.NET Core*Windows Server Hosting*' 1.1.30327.81))
 {
-    Write-Host ".NET Core Windows Server Hosting Bundle minimum version 1.1.30327.81 not installed...download and install from https://go.microsoft.com/fwlink/?linkid=844461. Halting installation."
-    exit 1
+   try{
+			Write-Host "Windows Server Hosting Bundle version 1.1.30327.81 not installed...installing version 1.1.30327.81"
+			Invoke-WebRequest -Uri https://go.microsoft.com/fwlink/?linkid=844461 -OutFile $env:Temp\bundle.exe
+			Start-Process $env:Temp\bundle.exe -Wait -ArgumentList '/quiet /install'
+			net stop was /y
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -102,10 +102,27 @@ if((Test-Path $zipPackage))
...
-    exit 1
+    try{
+		Write-Host "Windows Server Hosting Bundle version 1.1.30327.81 not installed...installing version 1.1.30327.81"
+        Write-Host "downloading to:" $env:Temp
+		Invoke-WebRequest -Uri https://go.microsoft.com/fwlink/?linkid=844461 -OutFile $env:Temp\bundle.exe
+		Start-Process $env:Temp\bundle.exe -Wait -ArgumentList '/quiet /install'
+		net stop was /y
+		net start w3svc			
+	}catch{
+	    Write-Error "Could not install .NET Windows Server Hosting bundle. Please install the hosting bundle before proceeding. https://go.microsoft.com/fwlink/?linkid=844461"
```
michaelvidal24: Do we know if this error will halt the script? Sometimes certain exceptions don't cause powershell to stop, and in this case we want the script to stop so the user can resolve the reported error.

Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -102,10 +102,27 @@ if((Test-Path $zipPackage))
 }
 
 
-if(!(Test-Prerequisite '*.NET Core*Windows Server Hosting*' 1.1.30327.81))
+if(!(Test-Prerequisite-Exact '*.NET Core*Windows Server Hosting*' 1.1.30327.81))
```
michaelvidal24: What happens if they have an earlier or later version installed? Are we able to halt the script and tell people what they need to do to resolve the issue?

kpaul: The script will fail with an error stating that a different version is installed. I should be able to add a message to the user to uninstall other versions.

Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
kpaul: The next line that throws the exception will halt the script. I can make the script a little smarter to let the user know if there are other versions installed and if that is not the cause of the error than just display the error message without throwing the exception. I will use ErrorAction -Stop to halt the script

PR #113, opened 2017-11-29
added code to unlock anonymous and windows auth config sections

kpaul:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/113)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -54,7 +54,13 @@ try{
 		Format-Table Id,Name,'Physical Path',Bindings -AutoSize
 
 	$selectedSiteId = Read-Host "Select a web site by Id:"
-	$selectedSite = $sites[$selectedSiteId - 1]
+
+	if($sites -is [array]){
+		$selectedSite = $sites[$selectedSiteId - 1]
+	}else{
+		$selectedSite = $sites
+	}
+
 	$webroot = $selectedSite.physicalPath
 	$siteName = $selectedSite.name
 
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -22,6 +22,22 @@ function Test-RegistrationComplete($authUrl)
     return $false
 }
 
+function Unlock-ConfigurationSections(){
+    $assembly = [System.Reflection.Assembly]::LoadFrom("$env:systemroot\system32\inetsrv\Microsoft.Web.Administration.dll")
```
michaelvidal24: We already import the powershell WebAdministration module, are we able to adjust these configurations using that powershell module directly?

Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -22,6 +22,22 @@ function Test-RegistrationComplete($authUrl)
...
+    
+    $section = $config.GetSection("system.webServer/security/authentication/anonymousAuthentication")
+    $section.OverrideMode = "Allow"    
+    Write-Host "Unlocked system.webServer/security/authentication/anonymousAuthentication"
+
+    $section = $config.GetSection("system.webServer/security/authentication/windowsAuthentication")
+    $section.OverrideMode = "Allow"    
+    Write-Host "Unlocked system.webServer/security/authentication/windowsAuthentication"
+    
+    $manager.CommitChanges()
```
michaelvidal24: What happens if the two sections are already unlocked, is an error thrown, or does it proceed silently? We would ultimately want to proceed silently since that wouldn't be a fatal error.

kpaul: if they are already unlocked it proceeds silently. no exception thrown

Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
kpaul: ahh yeah. i think i can just move it below the import of the utilities script

PR #120, opened 2017-12-01
Save the installation parameters back to install.config so they can b…

michaelvidal24:
…e used in subsequent installations.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/120)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -215,7 +215,7 @@ if(!(Test-PrerequisiteExact "*.NET Core*Windows Server Hosting*" 1.1.30327.81))
     Write-Success ".NET Core Windows Server Hosting Bundle installed and meets expectations."
 }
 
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -327,7 +328,14 @@ $body = @'
 
 Write-Console "Registering Fabric.Installer."
 $installerClientSecret = Add-ClientRegistration -authUrl $identityServerUrl -body $body
-Add-SecureInstallationSetting "common" "fabricInstallerSecret" $installerClientSecret $signingCert
+if($installerClientSecret){ Add-SecureInstallationSetting "common" "fabricInstallerSecret" $installerClientSecret $signingCert }
```
bsmithvt: Are the 2 lines below intentional (adding the same setting to "common" and "identity")?

michaelvidal24: Yes, they are intentional - probably overkill but we allow and individual setting to override the common setting.

PR #123, opened 2017-12-01
added ItemGroup section to publish files containing data and log moun…

bsmithvt:
…t points

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/123)


Fabric.Identity.SqlServer/Admin/MountPoints.sql
```diff
@@ -0,0 +1,20 @@
+﻿/*
+Do not change the database path or name variables.
+Any sqlcmd variables will be properly substituted during 
+build and deployment.
+*/
+
+ALTER DATABASE [$(DatabaseName)]
+    ADD FILE (NAME = [HCFabricIdentityData1File1], FILENAME = '$(FabricIdentityDataMountPoint)\HC$(DatabaseName)Data1File1.ndf', SIZE = 100 MB) TO FILEGROUP [HCFabricIdentityData1];
```
michaelvidal24: Let's check with some of our other teams to see if we should add a MAXSIZE and FILEGROWTH attributes here.

Fabric.Identity.SqlServer/Admin/MountPoints.sql
```diff
@@ -0,0 +1,20 @@
...
+
+GO
+
+ALTER DATABASE [$(DatabaseName)]
+    ADD LOG FILE (NAME = [HCFabricIdentityLogFile1], FILENAME = '$(FabricIdentityLogFileMountPoint)\HC$(DatabaseName)LogFile1.ldf', SIZE = 100 MB);
```
michaelvidal24: Same comment as above.

Fabric.Identity.SqlServer/Admin/MountPoints.sql
michaelvidal24: Do we also need an index file definition?

bsmithvt:
@scatmull, thanks for the feedback. We're using the same values as other database projects in the Core solution (IDEA, Metadata, etc.). Just want to confirm that we should indeed remove these defaults since they currently match existing databases.

scatmull:
You can leave it as is unless you know how to do it.  I looked at it briefly but could not find an obvious way to do it.

michaelvidal24:
@scatmull - To answer your question regarding concurrency, we run a performance test with 21 concurrent users with Identity installed on windows on top of a sql server database in our release pipeline which gets kicked off for each commit/merge to master.

PR #153, opened 2018-01-22
Ensure claims from 3rd party ID provider are pulled through.

michaelvidal24:
Fix up swagger URL when not running from the top level website.
Add config to disable Windows Auth when running in IIS.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/153)

### DIFF
--- a/Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
+++ b/Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
@@ -1,8 +1,9 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Claims;
 using System.Threading.Tasks;
-using Fabric.Identity.API.Persistence.SqlServer.EntityModels;
 using Fabric.Identity.API.Persistence.SqlServer.Services;
 using Microsoft.EntityFrameworkCore;
 using Fabric.Identity.API.Persistence.SqlServer.Mappers;
@@ -13,6 +14,7 @@ namespace Fabric.Identity.API.Persistence.SqlServer.Stores
     public class SqlServerUserStore : IUserStore
     {
         private readonly IIdentityDbContext _identityDbContext;
+        private static readonly ConcurrentDictionary<string, List<Claim>> UserClaims = new ConcurrentDictionary<string, List<Claim>>();
 
         public SqlServerUserStore(IIdentityDbContext identityDbContext)
         {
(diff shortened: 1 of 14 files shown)


michaelvidal24:
@kpaul, @bsmithvt - Can you guys take a look at my latest updates to this PR? I had to switch to persisting in the db to handle the scenario where the server restarted (thus clearing the ConcurrentDictionary) but the client still had a valid token and attempted to call the userinfo endpoint.

Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
```diff
@@ -66,9 +58,8 @@ public SqlServerUserStore(IIdentityDbContext identityDbContext)
         public async Task<User> AddUserAsync(User user)
         {
             var userEntity = user.ToEntity();
-
+            userEntity.Claims = user.Claims.Select(c => new UserClaim {Type = c.Type, Value = c.Value}).ToList();
```
kpaul: you could move this line into the ToEntity extension method that does not take an existing entity as a parameter.

Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
```diff
@@ -90,19 +81,21 @@ public void UpdateUser(User user)
...
+                if (userClaim != null)
+                {
+                    userClaim.Value = claim.Value;
+                }
+                else
+                {
+                    existingUser.Claims.Add(new UserClaim{Type = claim.Type, Value = claim.Value});
+                }
+            }
+
```
kpaul: you could move this line into the ToEntity extension method that accepts an existing entity as a parameter

Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
```diff
@@ -90,19 +81,21 @@ public void UpdateUser(User user)
 
             user.ToEntity(existingUser);
```
bsmithvt: Is there a case where claims can be removed?

michaelvidal24: Good point, i've updated the code to handle this situation, and also added tests to cover it.

Fabric.Identity.API/Persistence/SqlServer/Stores/SqlServerUserStore.cs
michaelvidal24: Good suggestions, I done that for both this comment and the one below.

bsmithvt:
Should Claim Tyoe and Value comparisons be case-insensitive?

michaelvidal24:
They should be case sensitive in this case as certain claims are case sensitive.

PR #155, opened 2018-02-08
User/bsmithvt/extension grant

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/155)

### DIFF
--- a/Fabric.Identity.API/Models/CustomGrantType.cs
+++ b/Fabric.Identity.API/Models/CustomGrantType.cs
@@ -1,0 +1,7 @@
+﻿namespace Fabric.Identity.API.Models
+{
+    public static class CustomGrantType
+    {
+        public const string Delegation = "delegation";
+    }
+}
--- a/Fabric.Identity.API/Models/CustomGrantType.cs
+++ b/Fabric.Identity.API/Models/CustomGrantType.cs
@@ -1,7 +1,0 @@
-﻿namespace Fabric.Identity.API.Models
-{
-    public static class CustomGrantType
-    {
-        public const string Delegation = "delegation";
-    }
-}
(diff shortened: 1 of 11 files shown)


Fabric.Identity.API/Models/CustomGrantType.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Fabric.Identity.API.Models
+{
+    public static class CustomGrantType
```
michaelvidal24: Is this an API model, or is it more of a constant? If its a constant, should we move it to the constants namespace?

Fabric.Identity.API/Validation/DelegationGrantValidator.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using IdentityModel;
+using IdentityServer4.Models;
+using IdentityServer4.Validation;
+
+namespace Fabric.Identity.API.Validation
```
michaelvidal24: Should we consider moving this to a new namespace called Fabric.Identity.API.ExtensionGrantValidators? While it is named *Validator, its not really synonymous with our other validators that validate API models.

Fabric.Identity.API/Models/CustomGrantType.cs
bsmithvt: Good point - IS4 put their GrantType.cs in a Models namespace, but I agree our Models namespaces has a different meaning. I'll move it.

Fabric.Identity.API/Validation/DelegationGrantValidator.cs
bsmithvt: Yes, I agree. Will move this as well.

PR #160, opened 2018-02-23
call the identity provider search service to get the user first and last name

kpaul:
added circuit breaker and flag to disable calling idp search

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/160)

### DIFF
--- a/Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
+++ b/Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
@@ -1,0 +1,84 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Fabric.Identity.API.Configuration;
+using Fabric.Identity.API.Models;
+using Fabric.Platform.Http;
+using IdentityModel.Client;
+using Newtonsoft.Json;
+using Serilog;
+
+namespace Fabric.Identity.API.Services
+{
+    public class IdPSearchServiceProvider : IExternalIdentityProviderService
+    {
+        private readonly IAppConfiguration _appConfig;
+        private readonly ILogger _logger;
+
+        public IdPSearchServiceProvider(IAppConfiguration appConfig, ILogger logger)
+        {
+            _appConfig = appConfig;
+            _logger = logger;
+        }
+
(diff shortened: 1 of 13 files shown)


Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+        private async Task<ExternalUser> SearchForUser(string subjectId)
+        {
+            var settings = _appConfig.IdentityServerConfidentialClientSettings;
+            var fabricIdentityClient = "fabric-identity-client";
+
+            var tokenUriAddress = $"{settings.Authority}connect/token";
+            var tokenClient = new TokenClient(tokenUriAddress, fabricIdentityClient, settings.ClientSecret);
+            var accessTokenResponse = await tokenClient.RequestClientCredentialsAsync("fabric/idprovider.searchusers");
+
+            var httpClient = new HttpClientFactory(
```
michaelvidal24: Do we need a using statement here so we dispose of the HttpClient properly?

Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            httpClient.DefaultRequestHeaders.Add("Accept", "application/json");
+
+            var searchServiceUrl = $"/v1/principals/user?subjectid={subjectId}";
+
+            var response = await httpClient.GetAsync(searchServiceUrl);
+
+            var responseContent = response.Content == null ? string.Empty : await response.Content.ReadAsStringAsync();
+
+            if (string.IsNullOrEmpty(responseContent) || response.StatusCode != HttpStatusCode.OK)
```
michaelvidal24: Should we log an error if the status code is InternalServerError?

Fabric.Identity.API/appsettings.json
```diff
@@ -1,9 +1,10 @@
 {
   "IssuerUri": "https://fabric.identity",
+
   "LogToFile": false,
   "ClientName": "hc",
   "RegistrationAdminGroup": "",
-  "AllowLocalLogin": "true",
+  "AllowLocalLogin": "false",
```
michaelvidal24: Lets revert the appsettings changes that we don't want to commit to master.

PR #175, opened 2018-03-20
do not require clientId in user search to support fabric.auth searche…

bsmithvt:
…s for users in roles that are shared across clients

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/175)

### DIFF
--- a/Fabric.Identity.API/Management/UsersController.cs
+++ b/Fabric.Identity.API/Management/UsersController.cs
@@ -102,13 +102,6 @@ namespace Fabric.Identity.API.Management
 
         private async Task<IActionResult> ProcessSearchRequest(string clientId, IEnumerable<string> userIds)
         {
-            if (string.IsNullOrEmpty(clientId))
-            {
-                return CreateFailureResponse(
-                    "No client id was included in the request. Please specify a client id",
-                    HttpStatusCode.BadRequest);
-            }
-
             var docIds = userIds.Select(id => id?.ToLower()).ToList();
             if (!docIds.Any() || docIds.All(string.IsNullOrEmpty))
             {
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/Management/UsersController.cs
```diff
@@ -102,26 +102,21 @@ public IActionResult Search(string searchText, string identityProvider)
...
+            if (!string.IsNullOrEmpty(clientId))
             {
-                return CreateFailureResponse($"The specified client with id: {clientId} was not found",
-                    HttpStatusCode.NotFound);
+                var client = _clientManagementStore.FindClientByIdAsync(clientId).Result;
+                if (string.IsNullOrEmpty(client?.ClientId))
+                {
+                    return CreateFailureResponse($"The specified client with id: {clientId} was not found",
+                        HttpStatusCode.NotFound);
+                }
```
michaelvidal24: What happens in line 124 when clientId is null or empty when we call ToUserViewModel?

bsmithvt: A null `LastLoginDate` is returned. Below is the implementation.

```
 public static UserApiModel ToUserViewModel(this User user, string clientId)
        {
            var lastLoginForClient = user.LastLoginDatesByClient
                .SingleOrDefault(l => l.ClientId.Equals(clientId, StringComparison.OrdinalIgnoreCase));

            DateTime? dateToSet = lastLoginForClient?.LoginDate;

            return new UserApiModel
            {
                SubjectId = user.SubjectId,
                FirstName = user.FirstName,
                LastName = user.LastName,
                MiddleName = user.MiddleName,
                LastLoginDate = dateToSet
            };
        }
```

bsmithvt: I'll add a check to the beginning of that method so we bypass all of the "last login date" logic if `clientId` is null or empty.

michaelvidal24: Thanks.

kpaul:
just a general comment, we will want to update the docs as well for this endpoint now that client id is not required

PR #180, opened 2018-03-23
fixed path concatenation in windows install script

benjanderson:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/180)


benjanderson:
This same fix can be made to Authorization install script as well. If you would like me to submit a new pull request I can

PR #188, opened 2018-04-03
Minor fixes to issues found during installation

michaelvidal24:
Ensure the authority URL has a trailing slash before using it to create the token url.
Add additional logging so we have better insight into calls to idpsearch.
Add decryption for the client secret.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/188)

### DIFF
--- a/Fabric.Identity.API/Extensions/UtilityExtensions.cs
+++ b/Fabric.Identity.API/Extensions/UtilityExtensions.cs
@@ -7,7 +7,7 @@ namespace Fabric.Identity.API.Extensions
 {
     public static class UtilityExtensions
     {
-        public static string FormatUrl(this string url)
+        public static string EnsureTrailingSlash(this string url)
         {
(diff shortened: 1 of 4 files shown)


Fabric.Identity.API/Extensions/UtilityExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Fabric.Identity.API.Extensions
+{
+    public static class UtilityExtensions
+    {
+        public static string FormatUrl(this string url)
```
kpaul: Minor comment but maybe change the name to something like EnsureTrailingSlash??

michaelvidal24: Good call, I'll update that.

PR #189, opened 2018-04-09
Adding script to upgrade admin roles from EDWAdmin to Authorization

petermcclanahan:
Adding an upgrade script to insert `EDW Admin` roles from `IdentityRoleBASE` into Authorization's `RoleUsers` table. This will be called at the end of the `Register.ps1` script at the end of the installer.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/189)

### DIFF
--- a/Fabric.Identity.API/scripts/Register.ps1
+++ b/Fabric.Identity.API/scripts/Register.ps1
@@ -684,4 +684,7 @@ Invoke-RegisterClients -clients $clients -identityServiceUrl $identityServiceUrl
 $authorization = Get-RolesAndPermissionsToRegister -registrationConfig $registrationSettings
 Invoke-RegisterSharedRolesAndPermissions -rolesAndPermissions $authorization -identityServiceUrl $identityServiceUrl -accessToken $accessToken
 
+# Upgrade users with EDWAdmin role to dosadmin role
+Invoke-Sqlcmd -InputFile "upgrade-admin-role.sql"
+
 Read-Host -Prompt "Registration complete, press Enter to exit"
--- a/Fabric.Identity.API/scripts/upgrade-admin-role.sql
+++ b/Fabric.Identity.API/scripts/upgrade-admin-role.sql
@@ -1,0 +1,46 @@
+  DECLARE @EdwAdminUsers TABLE
+  (
+	IdentityID INT NOT NULL,
+	IdentityNM VARCHAR(255) NOT NULL,
+	RoleNM VARCHAR(255) NOT NULL
+  )
+
+  DECLARE @DosAdminRoleId VARCHAR(255)
+
+
+BEGIN TRY
+	BEGIN TRANSACTION;
+
+	-- Get the members of IdentityBASE who have the 'EDW Admin' role
+	INSERT INTO @EdwAdminUsers
+	SELECT i.IdentityID, i.IdentityNM, r.RoleNM
+	FROM [EDWAdmin].[CatalystAdmin].[RoleBASE] r
+		INNER JOIN EDWAdmin.CatalystAdmin.IdentityRoleBASE ir ON r.RoleID = ir.RoleID
+		INNER JOIN EDWAdmin.CatalystAdmin.IdentityBASE i on ir.IdentityID = i.IdentityID
+	WHERE RoleNM = 'EDW Admin'
+
+	-- Insert the users from IdentityRoleBASE into dbo.Users that don't already exist in dbo.Users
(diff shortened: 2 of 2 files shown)


Fabric.Identity.API/scripts/upgrade-admin-role.sql
```diff
@@ -0,0 +1,46 @@
...
+
+	-- Get the dosadmin RoldId
+	SET @DosAdminRoleId = (SELECT RoleId from [Authorization].[dbo].[Roles] WHERE Name = 'dosadmin')
+
+	-- Insert these users into RoleUsers table associated with the dosadmin role
+	INSERT INTO [Authorization].[dbo].[RoleUsers] (CreatedBy, CreatedDateTimeUtc, RoleId, IdentityProvider, IsDeleted, SubjectId)
+	SELECT 'fabric-installer', getUtcDate(), @DosAdminRoleId, 'windows', 0, i.IdentityNM
+	FROM @EdwAdminUsers i
+
+	 SELECT * FROM [Authorization].[dbo].[RoleUsers]
```
michaelvidal24: Is there a reason we are selecting these after the insert is completed?

Fabric.Identity.API/scripts/Register.ps1
```diff
@@ -684,4 +684,7 @@ Invoke-RegisterClients -clients $clients -identityServiceUrl $identityServiceUrl
 $authorization = Get-RolesAndPermissionsToRegister -registrationConfig $registrationSettings
 Invoke-RegisterSharedRolesAndPermissions -rolesAndPermissions $authorization -identityServiceUrl $identityServiceUrl -accessToken $accessToken
 
-Read-Host -Prompt "Registration complete, press Enter to exit"
\ No newline at end of file
+# Upgrade users with EDWAdmin role to dosadmin role
+Invoke-Sqlcmd -InputFile "upgrade-admin-role.sql"
```
michaelvidal24: Is it possible to write a message to the console so the user running the script has some feedback as to what is happening?

Fabric.Identity.API/scripts/upgrade-admin-role.sql
```diff
@@ -0,0 +1,46 @@
...
+	  ON i.IdentityNM = u.SubjectId
+	WHERE u.SubjectId IS NULL
+
+	-- Get the dosadmin RoldId
+	SET @DosAdminRoleId = (SELECT RoleId from [Authorization].[dbo].[Roles] WHERE Name = 'dosadmin')
+
+	-- Insert these users into RoleUsers table associated with the dosadmin role
+	INSERT INTO [Authorization].[dbo].[RoleUsers] (CreatedBy, CreatedDateTimeUtc, RoleId, IdentityProvider, IsDeleted, SubjectId)
+	SELECT 'fabric-installer', getUtcDate(), @DosAdminRoleId, 'windows', 0, i.IdentityNM
+	FROM @EdwAdminUsers i
```
michaelvidal24: Should we also add a clause to make sure that the User who gets added during Authorization install isn't added here again? I think that would cause a unique constraint violation causing the script to fail.

Fabric.Identity.API/scripts/upgrade-admin-role.sql
```diff
@@ -0,0 +1,46 @@
...
+	LEFT JOIN [Authorization].[dbo].Users u
+	  ON i.IdentityNM = u.SubjectId
+	WHERE u.SubjectId IS NULL
+
+	-- Get the dosadmin RoldId
+	SET @DosAdminRoleId = (SELECT RoleId from [Authorization].[dbo].[Roles] WHERE Name = 'dosadmin')
+
+	-- Insert these users into RoleUsers table associated with the dosadmin role
+	INSERT INTO [Authorization].[dbo].[RoleUsers] (CreatedBy, CreatedDateTimeUtc, RoleId, IdentityProvider, IsDeleted, SubjectId)
```
michaelvidal24: RoleUsers associates Users to Roles, if any of these are AD groups the relationship should be made in GroupRoles.

Fabric.Identity.API/scripts/upgrade-admin-role.sql
petermcclanahan: Nope! Good catch. I'll remove that.

Fabric.Identity.API/scripts/Register.ps1
petermcclanahan: Definitely. Great suggestion.

michaelvidal24:
We changed approach on this item, so closing this PR.

PR #194, opened 2018-04-12
User/bsmithvt/remove web dav

bsmithvt:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/194)


Fabric.Identity.API/scripts/Register.ps1
```diff
@@ -435,6 +435,7 @@ function Invoke-WriteSecretToConfig($service, $secret, $configPath){
     
     if(!([string]::IsNullOrWhiteSpace($service.secretConfig))){
         Add-WebConfigAppSetting -webConfigLocation $configPath -settingKey $service.secretConfig -settingValue $secret | Out-Null
+		Write-Success "    Wrote secret to $service.serviceName configuration file as key=$service.secretConfig, value=$secret."
```
michaelvidal24: Should we add the path of the file to the log message also, we have it in $configPath I think.

bsmithvt: Yes - I meant to put that in. Thanks.

PR #211, opened 2018-05-22
Allow registration of hybrid/Pkce clients not in Discovery Service

MikeCronq:
This allows clients like SAMD which are not placed in the discovery service to bypass the discovery service request.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/211)


MikeCronq:
Added changes Mike Vidal suggested. Verified SAMD client was registered.

PR #212, opened 2018-06-04
User/jason/identityclientsecret

ctediz:
Check to see if fabricInstallerSecret was set, if not, register and create new installer secret.

Reset Identity client secret, had to cast and trim as well

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -563,7 +563,7 @@ if([string]::IsNullOrWhiteSpace($fabricInstallerSecret) -and [string]::IsNullOrE
 '@
 
     Write-Console "Registering Fabric.Installer Client."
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -549,19 +549,23 @@ $body = @'
 Write-Console "Registering Fabric.Identity Registration API."
 $registrationApiSecret = Add-ApiRegistration -authUrl $identityServerUrl -body $body -accessToken $accessToken
```
michaelvidal24: Do we need to update this call also, to cast to string and then trim?

ctediz: Api secrets were being returned as expected, for some reason only client secrets were acting differently. I don't think it's a bad idea to cast and trim them to be safe though.

PR #217, opened 2018-06-22
added upgrade change to append a /v1 to discoveryservice url if not present

hckenmiller:
I got an error upgrading, but not the same error, I got it in the DiscoveryServiceClient, so I added code to add /v1 to the default discovery service url, if not present from having a previous installation. 
I still added more error code to where his error was thrown in the Fabric.Registration, but I think the natural progression would be to upgrade first Fabric.Identity, then Fabric.Authorization, then Fabric.Registration, and this code addition will work. If they were to skip over Fabric.Identity, then the /v1 code addition would need to be duplicated in the others.  I haven't been able to test the error code addition, since I didn't get the same error as him. May need to over the shoulder with him, through his process before I check in.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/217)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
@@ -151,10 +151,10 @@ function Get-DiscoveryServiceUrl($discoUrl)
     if([string]::IsNullOrEmpty($discoUrl)){
         return "https://$env:computername.$($env:userdnsdomain.tolower())/DiscoveryService/v1"
     }else{
-	      $discoUrl = $discoUrl.TrimEnd('/')
+	      $discoUrl = $discoUrl.TrimEnd("/")
           if ($discoUrl -notmatch "/v\d")
 		  {
-	  		  return $discoUrl + '/v1'
+	  		  return $discoUrl + "/v1"
 		  }
 		  else 
 		  {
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/scripts/Install-Identity-Windows.ps1
```diff
@@ -151,7 +151,15 @@ function Get-DiscoveryServiceUrl($discoUrl)
     if([string]::IsNullOrEmpty($discoUrl)){
         return "https://$env:computername.$($env:userdnsdomain.tolower())/DiscoveryService/v1"
     }else{
-        return $discoUrl
+		  $hasversion = $discoUrl -match "/v\d"
```
ryanorbaker: should we be scrubbing a trailing "/" as well?   I think other places in the installer trap/fix that.

hckenmiller: there is only an issue when I was doing a little exploratory with having a trailing backslash in the following scenario. http://localhost/DiscoveryService/. This scenario  http://localhost/DiscoveryService/v1/ still works fine and doesn't append another v1.

Fabric.Identity.API/scripts/Register.ps1
```diff
@@ -24,7 +24,8 @@ function Get-ApplicationUrl($serviceName, $serviceVersion, $discoveryServiceUrl)
     $discoveryResponse = Invoke-RestMethod -Method Get -Uri $discoveryRequest -UseDefaultCredentials
     $serviceUrl = $discoveryResponse.value.ServiceUrl
     if([string]::IsNullOrWhiteSpace($serviceUrl)){
-        throw "The service $serviceName version $serviceVersion is not registered with the Discovery service. Make sure that this version of the service is registered w/ Discovery service before proceeding. Halting installation."
+	    Write-Error "There was an error registering the service $serviceName, using DiscoveryService url $discoveryRequest."
```
michaelvidal24: I think this error should say something about "There was an error getting the service registrations for $serviceName using DiscoveryService url $discoveryRequest. I don't think we are registering with discovery here, simply trying to retrieve the discovery registration.

PR #219, opened 2018-06-22
identity cli skeleton and pester unit tests

ctediz:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/219)


Fabric.Identity.API/scripts/identity-cli.psm1
```diff
@@ -0,0 +1,141 @@
...
+
+    .Example
+    Get-AccessToken -identityUrl "https://server/identity" -clientId "sample-client-id" -secret "SECrEtStrING" -scopes "offline_access"
+#>
+function Get-AccessToken {
+    param(
+        [Parameter(Mandatory=$True)] [Uri] $identityUrl,
+        [Parameter(Mandatory=$True)] [string] $clientId,
+        [Parameter(Mandatory=$True)] [string] $secret,
+        [Parameter(Mandatory=$True)] [string] $scope
```
ctediz: Made these required, prompts in the script if not provided, not sure if this is the approach to take here

Fabric.Identity.API/scripts/tests/identity-cli-tests.ps1
```diff
@@ -0,0 +1,59 @@
...
+
+                $response = Get-FabricInstallerAccessToken -identityUrl $mockUrl -secret "Secret"
+
+                $response | Should -Be "return token"
+            }
+        }
+
+        Context 'Invalid Request' {
+            It 'Should return an exception when invalid response' {
+                Mock -ModuleName identity-cli -CommandName Invoke-RestMethod { throw  New-Object -TypeName "System.Net.WebException" -ArgumentList "Error" }
```
ctediz: Couldn't find a to create an object outside of the mock setup, trying to set '$exc = New-Object -TypeName "System.Net.WebException" -ArgumentList "Error"' and throwing the variable here raised a different exception that was expected. 
Trying to instead return an object using the same method would return null instead of the object

ctediz:
Leaving this over the weekend, dev ops has a build running their pester tests: https://healthcatalyst.visualstudio.com/CAP/DevOps%20Infrastructure/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=796

Fabric.Identity.API/scripts/identity-cli.psm1
ryanorbaker: does making these required prevent negative testing?
Also, can't scope be null? (to get all available scopes?)

michaelvidal24: I think making the parameters mandatory (where applicable) is good practice, it will leverage the built in parameter validation so we don't have to build or test that. I agree, that we can make $scope optional, since not specifying it should return a token for all explicitly allowed scopes.

Fabric.Identity.API/scripts/identity-cli.psm1
```diff
@@ -0,0 +1,141 @@
...
+        client_secret = "$secret"
+    }
+
+    $accessTokenResponse = Invoke-RestMethod -Method Post -Uri $url -Body $body
+    return $accessTokenResponse.access_token
+}
+
+<#
+    .Synopsis
+    Attempts to get an access token the fabric installer client
```
michaelvidal24: I think we are missing a word here: "Attempts to get an access token `for` the fabric installer client"

Fabric.Identity.API/scripts/identity-cli.psm1
```diff
@@ -0,0 +1,141 @@
...
+    Get-AccessToken -identityUrl "https://server/identity"-secret "SECrEtStrING" 
+#>
+function Get-FabricInstallerAccessToken {
+    param(
+        [Parameter(Mandatory=$True)] [Uri] $identityUrl,
+        [Parameter(Mandatory=$True)] [string] $secret
+    )
+
+    $clientId = "fabric-installer"
+    $scope = "fabric/identity.manageresources"
```
michaelvidal24: We may want to specify all the scopes that we register the fabric-installer client with, that way this access token is valid to perform actions in Fabric.Authorization as well.  See https://github.com/HealthCatalyst/Fabric.Identity/blob/master/Fabric.Identity.API/scripts/Install-Identity-Windows.ps1#L543.

Fabric.Identity.API/scripts/tests/identity-cli-tests.ps1
```diff
@@ -0,0 +1,59 @@
+# Force re-import to pick up latest changes
+Import-Module ..\identity-cli.psm1 -Force
+
+describe 'Get-AccessToken' {
+    Context 'Unit Tests' {
+        Context 'Valid Request' {
+            It 'Should return an access token when valid response' {                
+                $mockUrl = New-MockObject -Type Uri
+                Mock -ModuleName identity-cli -CommandName Invoke-RestMethod { return @{
+                    access_token = "return token"
```
michaelvidal24: Should we mock the full response that the server would send back, to make sure that our code is parsing and returning the right value?

Fabric.Identity.API/scripts/identity-cli.psm1
```diff
@@ -0,0 +1,141 @@
...
+    .Parameter identityUrl
+    The identity url to get the access token from
+
+    .Parameter secret
+    The secret of the fabric installer client
+
+    .Parameter scopes
+    The scope of the access token to be requested
+
+    .Example
```
bsmithvt: Is there a space missing before `-secret` in this comment?

ctediz:
Going to complete this pr and create another for functional tests

PR #222, opened 2018-06-27
User/ken/registration bug changes

hckenmiller:
I was able to repro Matt's bug, now that I know the repro steps. Here is the fix for that.


Fabric.Identity.API/scripts/Register.ps1
```diff
@@ -24,8 +24,8 @@ function Get-ApplicationUrl($serviceName, $serviceVersion, $discoveryServiceUrl)
     $discoveryResponse = Invoke-RestMethod -Method Get -Uri $discoveryRequest -UseDefaultCredentials
     $serviceUrl = $discoveryResponse.value.ServiceUrl
     if([string]::IsNullOrWhiteSpace($serviceUrl)){
-	    Write-Error "There was an error getting the service registration for $serviceName, using DiscoveryService url $discoveryRequest."
```
ctediz: Where does the Write-Error write to, can it still exist and Write-Error $addToError as well?

hckenmiller: When I added the Write-Error, I was under the impression that it would output that error first and then the throw would output it's error. In testing the throw swallowed the write-error.

hc-jordenlowe: If you want to just display an error without throwing an exception, I believe you can give it another parameter

Nevermind, you have to give it that parameter to cause the script to halt.
Write-Error ""  -ErrorAction Stop

PR #223, opened 2018-06-27
User/jason/clifunctionaltests

ctediz:
Functional tests setup for the identity cli


ctediz:
This is in progress, sets up locally, working on getting the setup to run on build server

Fabric.Identity.API/scripts/tests/identity-cli.tests.ps1
```diff
@@ -71,4 +72,22 @@ describe 'Get-FabricInstallerAccessToken' {
             }
         }
     }
+}
+
+describe 'Functional-Tests' {    
+    BeforeAll {
+        # Set up auth/identity
+        Start-Process "$targetFilePath\..\start-auth-identity.sh" -Wait
```
ctediz: these launch a new process, so this script and the stop-auth-identity script may be failing on the build server which and the test continues regardless.  I imagine there's a way to check the status of the process, will need to look into it

Fabric.Identity.API/scripts/identity-cli.psm1
```diff
@@ -27,8 +27,10 @@ function Get-AccessToken {
         [Parameter(Mandatory=$True)] [string] $secret,
         [string[]] $scope
     )
+    
+    $url = ""
+    [System.Uri]::TryCreate($identityUrl, "connect/token", [ref]$url) | Out-Null
```
michaelvidal24: Do we want to do a TryCreate here? Should we just use the Uri constructor that takes a string and let that throw an exception if the Uri is in an invalid format?

Fabric.Identity.API/scripts/tests/identity-cli.functional.tests.ps1
```diff
@@ -0,0 +1,122 @@
...
+                }
+                catch {
+                    $error = Get-ErrorFromResponse -response $_.Exception.Response
+
+                    $_.Exception | Should -BeOfType System.Net.WebException 
+                    $error | Should -Be '{"error":"invalid_client"}'
+                }
+            }
+
+            It 'Should return an access token when valid request without a scope' {
```
michaelvidal24: Is this test named properly? Seems like we are expecting an error, not an access token...

Fabric.Identity.API/scripts/tests/identity-cli.functional.tests.ps1
```diff
@@ -0,0 +1,122 @@
...
+                }
+                catch {
+                    $_.Exception | Should -BeOfType System.Net.WebException
+                    $error = Get-ErrorFromResponse -response $_.Exception.Response
+
+                    $error | Should -Be '{"error":"invalid_scope"}'
+                }
+            }
+
+            It 'Should return an access token when valid request without a valid secret' {
```
michaelvidal24: Is this test named properly? Seems like we are expecting an error, not an access token...

Fabric.Identity.API/scripts/identity-cli.psm1
ctediz: Yes, this doesn't work as expected, this was a workaround for running against the docker identity, which breaks against a real url.  The issue there was that a trailing slash was being appended regardless, which was causing issues, going to work around it by pulling out the string from the uri and using that instead

PR #227, opened 2018-07-13
Refactored decryption logic out of IdentityConfigurationProvider and …

michaelvidal24:
…into a DecryptionService. Added unit and integration tests for decrypting encrypted data. Fixed bug where we were trimming improperly.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/227)

### DIFF
--- a/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
+++ b/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
@@ -1,6 +1,4 @@
-﻿using System.Security.Cryptography;
-using System.Security.Cryptography.X509Certificates;
-using Fabric.Identity.API.Services;
+﻿using Fabric.Identity.API.Services;
 using Fabric.Platform.Shared.Configuration.Docker;
 using Microsoft.Extensions.Configuration;
 
@@ -8,12 +6,10 @@ namespace Fabric.Identity.API.Configuration
 {
     public class IdentityConfigurationProvider
     {
-        private static readonly string EncryptionPrefix = "!!enc!!:";
-        
-        public IAppConfiguration GetAppConfiguration(string baseBath, ICertificateService certificateService)
+        public IAppConfiguration GetAppConfiguration(string baseBath, DecryptionService decryptionService)
         {
             var appConfig = BuildAppConfiguration(baseBath);
-            DecryptEncryptedValues(appConfig, certificateService);
+            DecryptEncryptedValues(appConfig, decryptionService);
             return appConfig;
         }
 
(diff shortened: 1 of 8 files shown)


Fabric.Identity.API/Services/LinuxCertificateService.cs
```diff
@@ -40,6 +41,12 @@ public X509Certificate2 GetEncryptionCertificate(SigningCertificateSettings cert
             throw new FabricConfigurationException("Do not encrypt settings when running on a Linux container, instead use Docker Secrets to protect sensitive configuration settings.");
         }
```
bsmithvt: I noticed this implementation is the same as the WindowsCertificateService. Will this always be the case? If so, maybe make this virtual in a base class as the default implementation?

Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
```diff
@@ -1,19 +1,15 @@
-﻿using System.Security.Cryptography;
-using System.Security.Cryptography.X509Certificates;
-using Fabric.Identity.API.Services;
+﻿using Fabric.Identity.API.Services;
```
hc-jordenlowe: Moving using directives inside the namespace scope would avoid naming issues.  However, because of the style of the application, and there are no conflicts, its probably save to keep these as is.

Fabric.Identity.API/Services/LinuxCertificateService.cs
bsmithvt: @michaelvidal24, not sure my suggestion is worth it for 2 lines of code?

Fabric.Identity.UnitTests/Services/DecryptionServiceTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            // Act
+            var decryptedString = decryptionService.DecryptString(clearTextstringToDecrypt, signingCertificateSettings);
+
+            // Assert
+            Assert.Equal(clearTextstringToDecrypt, decryptedString);
+        }
+
+        [Theory]
+        [InlineData(null)]
```
hc-jordenlowe: I like this part about XUnit vs MStest.  These InlineData allow you to compress 3 tests into one.  Cuts out a bunch of bloat.

Fabric.Identity.API/Services/LinuxCertificateService.cs
michaelvidal24: I had this exact debate before I pushed up the branch, :-). My initial thought was that the overhead of an abstract base class outweighed the benefit, since this is the only small method with overlap in implementation.

bsmithvt: Yeah, after thinking about it, I think I would just leave it as is also.

PR #229, opened 2018-07-13
User/ryano/c li client functions2

ryanorbaker:
for review, I still have merger to do

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/229)


Fabric.Identity.API/scripts/tests/CatalystDosIdentity.tests.ps1
```diff
@@ -0,0 +1,461 @@
+param(
+    [string] $targetFilePath = "$PSScriptRoot\..\CatalystDosIdentity.psm1"
+)
+
+# Force re-import to pick up latest changes
+Import-Module $targetFilePath -Force
+
+describe 'Get-AccessToken Unit Tests' -tag 'Unit' {
```
hckenmiller: Is there a reason you are formatting the Describe with a lower case D?

Fabric.Identity.API/scripts/tests/CatalystDosIdentity.tests.ps1
```diff
@@ -0,0 +1,461 @@
...
+    Context 'Invalid Request - existing client, invalid json' {
+        It 'New Registration should throw an exception when client already exists, and new json is invalid' {
+            $authUrl = [System.Uri]'https://some.Server/identity2'
+
+            $newClient = New-ClientCredentialsClientBody `
+                -clientId "cliTestClient" `
+                -clientName "cli Test Client Name" `
+                -allowedScopes @("fabric/identity.manageresources", "fabric/authorization.read", "fabric/authorization.write", "fabric/authorization.manageclients")
+
+            # $newClient.allowedGrantTypes = ""
```
hckenmiller: can this be removed

Fabric.Identity.API/scripts/tests/CatalystDosIdentity.tests.ps1
```diff
@@ -0,0 +1,461 @@
...
+}
+
+describe 'New-ClientCredentialsClientBody Unit Tests' -tag 'Unit' {
+    Context 'Valid Object' {
+        It 'New-ClientCredentialsClientBody should return a valid object with required parameters' {
+            $myClientId = "validClientCredentialsClient"
+            $myClientName = "Valid client credentials client name"
+
+            $response = New-ClientCredentialsClientBody -clientId $myClientId -clientName $myClientName -allowedScopes @("fabric/authorization.read", "dos/metadata", "fabric.profile")
+            $response | Should -not -Be :null
```
hckenmiller: doesn't affect function, but we probably should camel case Should -Not -Be, instead of Should -not -Be or Should -not -be

Fabric.Identity.API/scripts/tests/CatalystDosIdentity.tests.ps1
```diff
@@ -0,0 +1,461 @@
...
+        It 'Should return an error when bad token' {
+            $mockUrl = New-MockObject -Type Uri
+            Mock -ModuleName CatalystDosIdentity -CommandName Invoke-RestMethod { throw  New-Object -TypeName "System.Net.WebException" -ArgumentList "Exception" }
+
+            {Get-ClientRegistration -identityUrl $mockUrl -clientId "someClient" -accessToken  "Bearer badtoken"} | Should -Throw "Exception" -ExceptionType System.Net.WebException
+        }
+    }
+}
+
+describe 'New-ClientRegistrationTests Unit Tests' -tag 'Unit' {
```
hckenmiller: Remove Tests on the end of ClientRegistrationTests

Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
```diff
@@ -0,0 +1,620 @@
...
+    The scope of the access token to be requested
+
+    .Example
+    Get-AccessToken -identityUrl "https://server/identity" -clientId "sample-client-id" -secret "SECrEtStrING" -scopes "offline_access"
+#>
+function Get-AccessToken {
+    param(
+        [Parameter(Mandatory = $True)] [Uri] $identityUrl,
+        [Parameter(Mandatory = $True)] [string] $clientId,
+        [Parameter(Mandatory = $True)] [string] $secret,
```
michaelvidal24: I didn't catch this in the initial review - but perhaps this parameter should be securestring?

Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
```diff
@@ -0,0 +1,620 @@
...
+    .Parameter secret
+    The secret of the fabric installer client
+
+    .Example
+    Get-AccessToken -identityUrl "https://server/identity" -secret "SECrEtStrING"
+#>
+function Get-FabricInstallerAccessToken {
+    param(
+        [Parameter(Mandatory = $True)] [Uri] $identityUrl,
+        [Parameter(Mandatory = $True)] [string] $secret
```
michaelvidal24: Same comment as above, might need to be a securestring.

Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
```diff
@@ -0,0 +1,620 @@
...
+    The base identity url
+
+    .Parameter body
+    the json payload of attributes and values for the new client
+
+    .Parameter accessToken
+    an access token previously retrieved from the identity server
+
+    .Example
+    New-ClientRegistration -identityUrl "https://server/identity" -body @'{"clientId":"fabric-installer", "clientName":"Fabric Installer" ...... }@' -accessToken "eyJhbGciO"
```
ctediz: The @'s are for multiline strings right? I don't think those need to be here

ryanorbaker: @ are for structures, but in this case we are expecting a single json string, so still not necessary.  Although powershell will auto-convert the hasttable @{} into a json string for us.

ryanorbaker: you are correct, removed

ryanorbaker:
as to naming, I'd like to keep Fabric out of the name.   We should not be shipping our org chart, and from the the customer's perspective, DOS is the product.   
This is not a super strong opinion, so let's discuss if someone has a strong opinion to put fabric into the name,

michaelvidal24:
Two additional points about the name to counter your argument about org chart :-)

1) Fabric.Identity is the name of the service that this CLI works with.
2) The org name of Fabric is inherited from the fact that the name of the service layer in DOS is called Fabric.

@johnrulon - any strong opinions here?

PR #236, opened 2018-07-19
make body type more flexible for json or object

ryanorbaker:
removed the type requirement for Body so it can be json string, client object or hash table of properties

### DIFF
--- a/Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
+++ b/Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
@@ -127,11 +127,19 @@ function Get-ClientRegistration {
 function New-ClientRegistration {
     param(
         [Parameter(Mandatory = $True)] [Uri] $identityUrl,
-        [Parameter(Mandatory = $True)] [string] $body,
+        [Parameter(Mandatory = $True)] $body,
         [Parameter(Mandatory = $True)] [string] $accessToken
     )
 
     [Uri] $url = "$($identityUrl.OriginalString.TrimEnd("/"))/api/client"
+
+    if (!($body -is [string])) {
+        $clientObject = $body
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
```diff
@@ -127,12 +127,20 @@ function Get-ClientRegistration {
...
-        [Parameter(Mandatory = $True)] [string] $body,
+        [Parameter(Mandatory = $True)] $body,
         [Parameter(Mandatory = $True)] [string] $accessToken
     )
 
     [Uri] $url = "$($identityUrl.OriginalString.TrimEnd("/"))/api/client"
 
+    if (!($body -is [string])) {
+        $clientObject = $body
+        $body = $body | ConvertTo-Json
```
hckenmiller: Is it possible that ConvertTo-Json can throw an error. This is what I have seen can happen in forums. ConvertTo-Json : The converted JSON string is in bad format. I guess if that error happens it is pretty clear, it might not need to go threw a try catch.

Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
```diff
@@ -127,12 +127,20 @@ function Get-ClientRegistration {
...
+    }
+    else {
+        $clientObject = ConvertFrom-Json -InputObject $body
+    }
+
```
hckenmiller: I like how you consolidated even more with the if else

Fabric.Identity.API/scripts/CatalystDosIdentity.psm1
ryanorbaker: I would view that error as a parameter error, so I think it's ok to bypass the try/catch

PR #239, opened 2018-07-27
Installer client now updates on upgrades

ctediz:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/239)


ryanorbaker:
does this mean the installer client secret changes on every install?

ctediz:
No, the installer secret will still the same for each install, [https://github.com/HealthCatalyst/InstallScripts/pull/49](related change to not reset the installer secret)

PR #240, opened 2018-08-16
Update Register.ps1 to associate a role to a group based on config.

michaelvidal24:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/240)

### DIFF
--- a/Fabric.Identity.API/scripts/Register.ps1
+++ b/Fabric.Identity.API/scripts/Register.ps1
@@ -194,6 +194,7 @@ function Invoke-AddOrGetRole($authUrl, $name, $displayName, $description, $grain
 
 function Invoke-AddOrGetGroup($authUrl, $name, $source, $accessToken){
     try{
+        Write-Host "    Adding group $name..."
         $group = Add-Group -authUrl $authorizationServiceURL -name $name -source $source -accessToken $accessToken
         return $group
     }catch{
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Register.ps1
```diff
@@ -235,12 +260,36 @@ function Add-PermissionToRole($authUrl, $roleId, $permission, $accessToken)
...
+function Add-RoleToGroupSafe($authUrl, $groupName, $role, $accessToken){
+    try{
+        Write-Host "    Adding Role: $($role.name) to Group $groupName..."
+        Add-RoleToGroup -authUrl $authUrl -groupName $groupName -role $role -accessToken $accessToken
+    }catch{
+        $exception = $_.Exception
+        if($exception -ne $null -and $exception.Response -ne $null)
+        {
+            $error = Get-ErrorFromResponse -response $exception.Response
+            if($error.Contains("$($role.id) already exists")){
```
bsmithvt: Do we want this check or a 409 check in case error message changes?

michaelvidal24: We have to first refactor the API to return a 409, currently it returns a 400 in this case unfortunately. We can open an issue in the Fabric.Authorization repo for that work, then we can update this.

PR #245, opened 2018-08-28
User/mvidal/install identity v2

michaelvidal24:
Adding new PowerShell script that enables quiet installs of Identity.
Uses a new PowerShell Module that has unit tests (60% coverage, as portions of the module are not worth unit testing as they are simply passthroughs to database or rest calls.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/245)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -155,8 +155,11 @@ function Get-Certificates([string] $primarySigningCertificateThumbprint, [string
     return @{SigningCertificate = $signingCert; EncryptionCertificate = $encryptionCert}
 }
 
-function Get-IISAppPoolUser([string] $appName, [string] $storedIisUser){
-    if(Test-AppPoolExistsAndRunsAsUser -appPoolName $appName -userName $storedIisUser){
+function Get-IISAppPoolUser([PSCredential] $credential, [string] $appName, [string] $storedIisUser){
+    if($credential){
+        $iisUser = "$($credential.GetNetworkCredential().Domain)\$($credential.GetNetworkCredential().UserName)"
+    }
+    elseif(Test-AppPoolExistsAndRunsAsUser -appPoolName $appName -userName $storedIisUser){
         $iisUser = $storedIisUser
     }
     else{
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -132,8 +132,6 @@ function Get-Certificates([string] $primarySigningCertificateThumbprint, [string
             }
     
         }catch{
-            $scriptDirectory =  Get-CurrentScriptDirectory
-            Set-Location $scriptDirectory
             Write-DosMessage -Level "Error" -Message  "Could not set the certificate thumbprint. Error $($_.Exception.Message)"
             throw
         }
@@ -175,15 +173,7 @@ function Get-IISAppPoolUser([string] $appName, [string] $storedIisUser){
         
             $iisUser = $userEnteredIisUser
             $userEnteredPassword = Read-Host "Enter the password for $iisUser" -AsSecureString
-            $credential = New-Object -TypeName "System.Management.Automation.PSCredential" -ArgumentList $iisUser, $userEnteredPassword
-            [System.Reflection.Assembly]::LoadWithPartialName("System.DirectoryServices.AccountManagement") | Out-Null
-            $ct = [System.DirectoryServices.AccountManagement.ContextType]::Domain
-            $pc = New-Object System.DirectoryServices.AccountManagement.PrincipalContext -ArgumentList $ct,$credential.GetNetworkCredential().Domain
-            $isValid = $pc.ValidateCredentials($credential.GetNetworkCredential().UserName, $credential.GetNetworkCredential().Password, [System.DirectoryServices.AccountManagement.ContextOptions]::Negotiate)
-            if(!$isValid){
-                Write-DosMessage -Level "Error" -Message "Incorrect credentials for $iisUser"
-                throw
-            }
+            $credential = Confirm-Credentials -iisUser $iisUser -userEnteredPassword $userEnteredPassword
             Write-DosMessage -Level "Information" -Message "Credentials are valid for user $iisUser"
         }else{
             Write-DosMessage -Level "Error" -Message "No user account was entered, please enter a valid user account."
@@ -192,6 +182,20 @@ function Get-IISAppPoolUser([string] $appName, [string] $storedIisUser){
     }
     if($iisUser){ Add-InstallationSetting "identity" "iisUser" "$iisUser" | Out-Null }
     return @{UserName = $iisUser; Credential = $credential}
+}
+
+function Confirm-Credentials([string] $iisUser, [SecureString] $userEnteredPassword){
+    $credential = New-Object -TypeName "System.Management.Automation.PSCredential" -ArgumentList $iisUser, $userEnteredPassword
+    [System.Reflection.Assembly]::LoadWithPartialName("System.DirectoryServices.AccountManagement") | Out-Null
+    $ct = [System.DirectoryServices.AccountManagement.ContextType]::Domain
+    $pc = New-Object System.DirectoryServices.AccountManagement.PrincipalContext -ArgumentList $ct,$credential.GetNetworkCredential().Domain
+    Write-Host "Confirming credentials"
+    $isValid = $pc.ValidateCredentials($credential.GetNetworkCredential().UserName, $credential.GetNetworkCredential().Password, [System.DirectoryServices.AccountManagement.ContextOptions]::Negotiate)
+    if(!$isValid){
+        Write-DosMessage -Level "Error" -Message "Incorrect credentials for $iisUser"
+        throw
+    }
+    return $credential
 }
 
 function Add-PermissionToPrivateKey($iisUser, $signingCert, $permission){
@@ -278,7 +282,7 @@ function Get-DiscoveryServiceUrl([string]$discoveryServiceUrl, [bool]$quiet){
     if(!$quiet){
         $userEnteredDiscoveryServiceUrl = Read-Host "Press Enter to accept the default DiscoveryService URL [$defaultDiscoUrl] or enter a new URL"
         if(![string]::IsNullOrEmpty($userEnteredDiscoveryServiceUrl)){   
-            $discoveryServiceUrl = $userEnteredDiscoveryServiceUrl
+            $defaultDiscoUrl = $userEnteredDiscoveryServiceUrl
         }
     }
     if($defaultDiscoUrl){ Add-InstallationSetting "common" "discoveryService" "$defaultDiscoUrl" | Out-Null }
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -6,15 +6,24 @@ if (!(Test-Path $fabricInstallUtilities -PathType Leaf)) {
 }
 Import-Module -Name $fabricInstallUtilities -Force
 
+# Import Dos Install Utilities
+$minVersion = [System.Version]::new(1, 0, 164 , 0)
 $dosInstallUtilities = Get-Childitem -Path ./**/DosInstallUtilities.psm1 -Recurse
 if ($dosInstallUtilities.length -eq 0) {
-    Install-Module DosInstallUtilities -Scope CurrentUser
-    Import-Module DosInstallUtilities -Force
-    Write-DosMessage -Level "Warning" -Message "Could not find dos install utilities. Manually installing"
+    $installed = Get-Module -Name DosInstallUtilities
+    if ($null -eq $installed) {
+        $installed = Get-InstalledModule -Name DosInstallUtilities
+    }
+
+    if (($null -eq $installed) -or ($installed.Version.CompareTo($minVersion) -lt 0)) {
+        Write-Host "Installing DosInstallUtilities from Powershell Gallery"
+        Install-Module DosInstallUtilities -Scope CurrentUser -MinimumVersion 1.0.164.0 -Force
+        Import-Module DosInstallUtilities -Force
+    }
 }
 else {
+    Write-Host "Installing DosInstallUtilities at $($dosInstallUtilities.FullName)"
     Import-Module -Name $dosInstallUtilities.FullName
-    Write-DosMessage -Level "Verbose" -Message "Installing DosInstallUtilities at $($dosInstallUtilities.FullName)"
 }
 
 function Get-FullyQualifiedInstallationZipFile([string] $zipPackage, [string] $workingDirectory){
@@ -49,7 +58,7 @@ function Install-DotNetCoreIfNeeded([string] $version, [string] $downloadUrl){
             Remove-Item $env:Temp\bundle.exe
         }catch{
             $e = $_.Exception
-            Write-DosMessage -Level "Warning" -Message "Unable to remove temporary dowload file for server hosting bundle exe" 
+            Write-DosMessage -Level "Warning" -Message "Unable to remove temporary download file for server hosting bundle exe" 
             Write-DosMessage -Level "Warning" -Message  $e.Message
         }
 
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,621 @@
...
+
+        return $selectedSite
+
+    }catch{
+        Write-DosMessage -Level "Error" -Message "Could not select a website."
+        throw
+    }
+}
+
+function Get-Certificates([string] $primarySigningCertificateThumbprint, [string] $encryptionCertificateThumbprint, [bool] $quiet){
```
imranq2: should we filter this to only SSL certs?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,621 @@
...
+        if(![string]::IsNullOrEmpty($userEnteredAppInsightsInstrumentationKey)){   
+            $appInsightsInstrumentationKey = $userEnteredAppInsightsInstrumentationKey
+        }
+    }
+    if($appInsightsInstrumentationKey){ Add-InstallationSetting "identity" "appInsightsInstrumentationKey" "$appInsightsInstrumentationKey" | Out-Null }
+    if($appInsightsInstrumentationKey){ Add-InstallationSetting "common" "appInsightsInstrumentationKey" "$appInsightsInstrumentationKey" | Out-Null }
+    return $appInsightsInstrumentationKey
+}
+
+function Get-SqlServerAddress([string] $sqlServerAddress, [bool] $quiet){
```
imranq2: should we default to current machine if no sql server is specified.  still should prompt user.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,621 @@
...
+                'Index'=$index;
+                'Subject'= $_.Subject; 
+                'Name' = $_.FriendlyName; 
+                'Thumbprint' = $_.Thumbprint; 
+                'Expiration' = $_.NotAfter
+                };
+                $index ++} |
+                Format-Table Index,Name,Subject,Expiration,Thumbprint  -AutoSize | Out-Host
+    
+            $selectionNumber = Read-Host  "Select a signing and encryption certificate by Index"
```
imranq2: instead of throwing an error why not have a while loop until they provide a value

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: Yes, we can look at ways to filter this list down.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: Good idea, I'll make that change.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: I believe we set localhost as the default in the install.config, and we go with that if nothing is entered.

michaelvidal24:
I may chunk the utitlities file into two different modules, since a lot of the functionality will be shared by Fabric.Authorization install also.

As far as the net stop stuff, we do check that the script is being run as administrator, and halt the script if its not being run as admin.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+if (!(Test-Path $fabricInstallUtilities -PathType Leaf)) {
+    Write-DosMessage -Level "Warning" -Message "Could not find fabric install utilities. Manually downloading and installing"
+    Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -Headers @{"Cache-Control" = "no-cache"} -OutFile $fabricInstallUtilities
+}
+Import-Module -Name $fabricInstallUtilities -Force
+
+$dosInstallUtilities = Get-Childitem -Path ./**/DosInstallUtilities.psm1 -Recurse
+if ($dosInstallUtilities.length -eq 0) {
+    Install-Module DosInstallUtilities -Scope CurrentUser
+    Import-Module DosInstallUtilities -Force
```
hckenmiller: I had been using force for the dbatools that is used by DosInstallUtilities. I had an issue with folders created for each version change, and that led to a 10 level deep folder nesting error. Since force retains all installed version changes in folders, if DosInstallUtilities has a lot of version changes we may need to clean out the folders or not use force. Just a thought.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+            Restart-W3SVC
+        }catch{
+            Write-DosMessage -Level "Error" -Message "Could not install .NET Windows Server Hosting bundle. Please install the hosting bundle before proceeding. $downloadUrl"
+            throw
+        }
+        try{
+            Remove-Item $env:Temp\bundle.exe
+        }catch{
+            $e = $_.Exception
+            Write-DosMessage -Level "Warning" -Message "Unable to remove temporary dowload file for server hosting bundle exe"
```
hckenmiller: download

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -0,0 +1,83 @@
...
+
+Test-MeetsMinimumRequiredPowerShellVerion -majorVersion 4
+
+if(!(Test-IsRunAsAdministrator))
+{
+    Write-DosMessage -Level "Error" -Message "You must run this script as an administrator. Halting configuration."
+    throw
+}
+
+$installSettings = Get-InstallationSettings "identity"
```
ctediz: this is much nicer than what we had before here

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
```diff
@@ -0,0 +1,761 @@
+param(
+    [string] $targetFilePath = "$PSScriptRoot\..\Install-Identity-Utilities.psm1"
+)
+
+Write-Host $targetFilePath
+# Force re-import to pick up latest changes
+Import-Module $targetFilePath -Force
+#Import-Module Pester -Force
```
ctediz: left over comment?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+                        'Id'=$_.id;
+                        'Name'=$_.name;
+                        'Physical Path'=[System.Environment]::ExpandEnvironmentVariables($_.physicalPath);
+                        'Bindings'=$_.bindings;
+                    };} |
+                    Format-Table Id,Name,'Physical Path',Bindings -AutoSize | Out-Host
+
+                do {
+                    $selectedSiteId = Read-Host "Select a web site by Id"
+                    $selectedSite = $sites[$selectedSiteId - 1]
```
bsmithvt: Should another try/catch be embedded here along with a loop to re-prompt user if `selectedSiteId` is out of range?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+function Get-IdentityDatabaseConnectionString([string] $identityDbName, [string] $sqlServerAddress, [bool] $quiet){
+    if(!$quiet){
+        $userEnteredIdentityDbName = Read-Host "Press Enter to accept the default Identity DB Name '$($identityDbName)' or enter a new Identity DB Name"
+        if(![string]::IsNullOrEmpty($userEnteredIdentityDbName)){
+            $identityDbName = $userEnteredIdentityDbName
+        }
+    }
+    $identityDbConnStr = "Server=$($sqlServerAddress);Database=$($identityDbName);Trusted_Connection=True;MultipleActiveResultSets=True;"
+
+    Invoke-Sql $identityDbConnStr "SELECT TOP 1 ClientId FROM Clients" | Out-Null
```
hckenmiller: If the DbName is changed and they don't accept the default. If they type it in wrong Database the doesn't exist, will there be a failure on the Invoke-Sql that is not caught?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+            do {
+                if($attempts -gt 10){
+                    Write-DosMessage -Level "Error" -Message "An invalid certificate has been selected."
+                    throw
+                }
+                $selectionNumber = Read-Host  "Select a signing and encryption certificate by Index"
+                if([string]::IsNullOrEmpty($selectionNumber)){
+                    Write-DosMessage -Level "Information" -Message "You must select a certificate so Fabric.Identity can sign access and identity tokens."
+                }else{
+                    $selectionNumberAsInt = [convert]::ToInt32($selectionNumber, 10)
```
bsmithvt: What happens here if if user enters non-numeric characters?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -0,0 +1,83 @@
...
+    -identityClientSecret $identityClientSecret `
+    -registrationApiSecret $registrationApiSecret `
+    -appDirectory $installApplication.applicationDirectory
+
+if ($fabricInstallerSecret){
+    Write-DosMessage -Level "Information" -Message "Please keep the following Fabric.Installer secret in a secure place, it will be needed in subsequent installations:"
+    Write-DosMessage -Level "Information" -Message "Fabric.Installer clientSecret: $fabricInstallerSecret"
+}
+
+if(!$quiet){
```
bsmithvt: If run in `quiet` mode, will this exit before the user has had a chance to write down secrets if they want to? If so, do we want that? I know the secrets are written to the appropriate app config files, but if something happens to go wrong there and the user needs them, what will be the repercussions? They will just need to re-run the script (potentially in non-quiet mode)?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: Invoke-Sql internally has a try catch that will catch the error, write out the error, and then forcibly stop the script using the `-ErrorAction Stop` parameter. The Out-Null is simply used here to suppress extraneous output from the successful execution of the sql statement.

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
```diff
@@ -0,0 +1,761 @@
...
+                $shouldShow = Test-ShouldShowCertMenu -primarySigningCertificateThumbprint "12345" -encryptionCertificateThumbprint "" -quiet $true
+                $shouldShow | Should -Be $false
+            }
+            It 'Should return false when in quiet mode and no thumbprints are present'{
+                $shouldShow = Test-ShouldShowCertMenu -primarySigningCertificateThumbprint "" -encryptionCertificateThumbprint "" -quiet $true
+                $shouldShow | Should -Be $false
+            }
+        }
+    }
+}
```
ctediz: We discussed writing tests for the installer at some point in the past, it's nice finally having some written though :)

Fabric.Identity.API/scripts/Install-Identity.ps1
michaelvidal24: It will, but the user will have the option of redirecting output to a file if they choose. There is also another workaround, the user can reset the installer secret if needed.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,642 @@
...
+                'Index'=$index;
+                'Subject'= $_.Subject; 
+                'Name' = $_.FriendlyName; 
+                'Thumbprint' = $_.Thumbprint; 
+                'Expiration' = $_.NotAfter
+                };
+                $index ++} |
+                Format-Table Index,Name,Subject,Expiration,Thumbprint  -AutoSize | Out-Host
+            do {
+                if($attempts -gt 10){
```
ctediz: Why does this function have an attempt counter?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: An exception will be thrown and the script will stop.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: I can add another check for that in the while.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: So that I can test it in pester without getting into an infinite loop :-).

michaelvidal24: I haven't figured out how to create a mock that does something different the second time the mock is called with the same parameters...

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
```diff
@@ -0,0 +1,761 @@
...
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName Get-FullyQualifiedMachineName -Times 1 -Exactly
+            }
+        }
+    }
+}
+
+Describe 'Test-ShouldShowCertMenu'{
+    InModuleScope Install-Identity-Utilities{
+        Context 'Interactive Mode'{
+            It 'Should return true when in interactive mode and signing cert thumprint is not present'{
```
hckenmiller: thumbprint

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
```diff
@@ -0,0 +1,761 @@
...
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName New-AppRoot -Times 1 -Exactly
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName New-App -Times 1 -Exactly
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName Publish-WebSite -Times 1 -Exactly
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName New-AppPool -Times 0 -Exactly
+
+            }
+        }
+    }
+
+    Context 'App pool does not exists'{
```
hckenmiller: should this be App pool exists?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
michaelvidal24: I'll update this so it only installs if we don't have the version we want, I think that should take care of this potential issue.

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
michaelvidal24: I think this is correct, if the app pool doesn't already exists, we are verifying that we do call the new app pool function. The above context tests the reverse, we validate that if the app pool does exist that we do call the new-apppool function.

Fabric.Identity.API/scripts/tests/Install-Identity-Utilities.tests.ps1
```diff
@@ -0,0 +1,761 @@
...
+                $shouldShow = Test-ShouldShowCertMenu -primarySigningCertificateThumbprint "12345" -encryptionCertificateThumbprint "12345" -quiet $false
+                $shouldShow | Should -Be $false
+            }
+            It 'Should return true when in interactive mode and no thumbprints are present'{
+                $shouldShow = Test-ShouldShowCertMenu -primarySigningCertificateThumbprint "" -encryptionCertificateThumbprint "" -quiet $false
+                $shouldShow | Should -Be $true
+            }
+        }
+        Context 'Quiet Mode'{
+            It 'Should return false when in quiet mode and signing and encryption cert thumbprnts are present'{
```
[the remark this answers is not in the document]
michaelvidal24: What we were trying to test here is basically that the cert selection menu never shows if we are in quiet mode, so for all of the tests in the 'Quiet Mode' context, we want to ensure that the Test-ShowShowCertMenu returns $false.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -0,0 +1,660 @@
+# Import Fabric Install Utilities
+$fabricInstallUtilities = ".\Fabric-Install-Utilities.psm1"
+if (!(Test-Path $fabricInstallUtilities -PathType Leaf)) {
+    Write-DosMessage -Level "Warning" -Message "Could not find fabric install utilities. Manually downloading and installing"
+    Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -Headers @{"Cache-Control" = "no-cache"} -OutFile $fabricInstallUtilities
+}
+Import-Module -Name $fabricInstallUtilities -Force
```
hckenmiller: I thought I commented here. In most of the places like above and below you are using force for import-module. Is it possible to have the 10 folder level version nesting issue, like with Install-Module when using force. Since all versions are saved locally.

michaelvidal24: You did comment, but on the dos install utilities one which I updated so github obsoleted the conversation. We aren't at risk for that with the fabric install utilities since it is simply a file w/ no folder structure that we download, then import.

PR #251, opened 2018-10-10
User/ryano/azure integration

ryanorbaker:
1. Changes to add Azure configuration.
2. Changes to allow local configuration file called: appsettings.Development.json.   Values in this file will NOT be added to the git repo, and supersede the values in the default appsettings.json file.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/251)

### DIFF
--- a/Fabric.Identity.API/Configuration/IAppConfiguration.cs
+++ b/Fabric.Identity.API/Configuration/IAppConfiguration.cs
@@ -4,8 +4,6 @@ using Fabric.Platform.Shared.Configuration;
 
 namespace Fabric.Identity.API.Configuration
 {
-    using System.Collections.Generic;
-
     public interface IAppConfiguration
     {
         string IssuerUri { get; }
--- a/Fabric.Identity.API/Configuration/IAppConfiguration.cs
+++ b/Fabric.Identity.API/Configuration/IAppConfiguration.cs
@@ -4,6 +4,8 @@ using Fabric.Platform.Shared.Configuration;
 
 namespace Fabric.Identity.API.Configuration
 {
+    using System.Collections.Generic;
+
     public interface IAppConfiguration
     {
         string IssuerUri { get; }
@@ -17,6 +19,8 @@ namespace Fabric.Identity.API.Configuration
 
         string DiscoveryServiceEndpoint { get; }
 
+        IEnumerable<string> ValidIssueres { get; }
+            
         IdentityProviderSearchSettings IdentityProviderSearchSettings { get; }
         SigningCertificateSettings SigningCertificateSettings { get; }
         ElasticSearchSettings ElasticSearchSettings { get; }
--- a/Fabric.Identity.API/Configuration/IAppConfiguration.cs
+++ b/Fabric.Identity.API/Configuration/IAppConfiguration.cs
@@ -14,19 +14,18 @@ namespace Fabric.Identity.API.Configuration
         string RegistrationAdminGroup { get; }
         bool AllowLocalLogin { get; }
         bool WindowsAuthenticationEnabled { get; }
-
+        bool AzureAuthenticationEnabled { get; }
         bool UseDiscoveryService { get; }
 
         string DiscoveryServiceEndpoint { get; }
 
-        IEnumerable<string> ValidIssueres { get; }
-            
         IdentityProviderSearchSettings IdentityProviderSearchSettings { get; }
         SigningCertificateSettings SigningCertificateSettings { get; }
         ElasticSearchSettings ElasticSearchSettings { get; }
         HostingOptions HostingOptions { get; }
         CouchDbSettings CouchDbSettings { get; }
         ExternalIdProviderSettings ExternalIdProviderSettings { get; }
+        AzureActiveDirectorySettings AzureActiveDirectorySettings { get; }
         IdentityServerConfidentialClientSettings IdentityServerConfidentialClientSettings { get; }
         ApplicationInsights ApplicationInsights { get; }
         LdapSettings LdapSettings { get; }
--- a/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
+++ b/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
@@ -4,24 +4,26 @@ using Microsoft.Extensions.Configuration;
 
 namespace Fabric.Identity.API.Configuration
 {
+
     public class IdentityConfigurationProvider
     {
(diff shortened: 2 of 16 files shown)


Fabric.Identity.API/Configuration/IAppConfiguration.cs
```diff
@@ -4,6 +4,8 @@
 
 namespace Fabric.Identity.API.Configuration
 {
+    using System.Collections.Generic;
```
michaelvidal24: Do we need this using?

Fabric.Identity.API/Quickstart/Account/AccountController.cs
```diff
@@ -237,9 +242,27 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl)
...
+
             if (userIdClaim == null)
             {
-                throw new Exception("Unknown userid");
+                throw new MissingUserClaimException(ExceptionMessageResources.MissingUserClaimMessage);
+            }
+
+            var schemaItem = info.Properties.Items.FirstOrDefault(i => i.Key == "scheme");
+
+            if (_appConfiguration.AzureAuthenticationEnabled && schemaItem.Value == FabricIdentityConstants.AuthenticationSchemes.Azure)
```
michaelvidal24: Do we need to check schemaItem for null here?

Fabric.Identity.API/Services/UserProfileService.cs
```diff
@@ -54,7 +60,16 @@ public async Task IsActiveAsync(IsActiveContext context)
             var sub = context.Subject.GetSubjectId();
             _logger.Debug($"found sub from IsActiveContext: {sub}");
             var user = await _userStore.FindBySubjectIdAsync(sub);
+
             context.IsActive = user != null;
+
+            if (user != null && _appConfig.AzureAuthenticationEnabled)
```
michaelvidal24: Do we also have to check the scheme that the user logged in with? If we have Windows auth enabled and Azure auth enabled, we might set the user to inactive if they are a windows user and we don't have an issuer for them.

Fabric.Identity.API/Startup.cs
```diff
@@ -208,6 +209,11 @@ public void ConfigureServices(IServiceCollection services)
             app.UseCors(FabricIdentityConstants.FabricCorsPolicyName);
 
             app.UseIdentityServer();
+            if (_appConfig.AzureAuthenticationEnabled)
```
michaelvidal24: Could we encapsulate this check in the UseAzureIdentityProvider function? We are passing in the config there anyway - might help to keep this already messy class a little cleaner.

Fabric.Identity.API/appsettings.json
```diff
@@ -42,6 +43,7 @@
   "ExternalIdProviderSettings": {
     "ExternalIdProviders": []
   },
+  "AzureActiveDirectorySettings": {},
```
michaelvidal24: Do we want to add some default settings here, so we can add comments for what the settings are for?

Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
```diff
@@ -4,24 +4,26 @@
...
-            return BuildAppConfiguration(basePath);
+            return BuildAppConfiguration(basePath, string.Empty);
         }
 
-        private IAppConfiguration BuildAppConfiguration(string baseBath)
+        private IAppConfiguration BuildAppConfiguration(string baseBath, string environmentName)
         {
                 var config = new ConfigurationBuilder()
                     .AddJsonFile("appsettings.json")
+                    .AddJsonFile($"appsettings.{environmentName}.json", true)
```
michaelvidal24: I Like this addition!

Fabric.Identity.API/Configuration/IAppConfiguration.cs
ryanorbaker: nope, it's a hangover.

Fabric.Identity.API/Quickstart/Account/AccountController.cs
ryanorbaker: nope.  schemaItem is actually a key/value pair.  A dictionary returns a "","" pair if first or default is not found.

Fabric.Identity.API/Startup.cs
ryanorbaker: ha, yes.   I was thinking the startup was big to make it more straightforward to read :-)

Fabric.Identity.API/Services/UserProfileService.cs
ryanorbaker: good point.  I wonder how this is working!

Fabric.Identity.API/appsettings.json
ryanorbaker: done

Fabric.Identity.API/Services/UserProfileService.cs
ryanorbaker: updated

Fabric.Identity.API/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -5,14 +5,56 @@
...
 
 namespace Fabric.Identity.API.Extensions
 {
+    using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+
     public static class ApplicationBuilderExtensions
     {
+
+
+        public static IApplicationBuilder UseAzureIdentityProviderIfAccplicable(this IApplicationBuilder builder,
```
michaelvidal24: Check spelling here: UseAzureIdentityProviderIfApplicable.

Fabric.Identity.API/Fabric.Identity.API.csproj
```diff
@@ -32,6 +32,10 @@
     <None Remove="Persistence\SqlServer\Entities\**" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Remove="appsettings.Development.json" />
```
hc-jordenlowe: is this file described somewhere?  Is that the documentation you are creating?

ryanorbaker: this is a copy file of the appsettings.json but it take precedence.   so you can have bogus values in appsettings and then create this file locally with your real secrets.

hc-jordenlowe: ok

PR #255, opened 2018-10-30
User/jason/idpssinstall

ctediz:
Workflow of Registering an application with Azure as well as a script to alter the config for idpss

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/255)


Fabric.Identity.API/scripts/Alter-Config.ps1
```diff
@@ -0,0 +1,47 @@
+# This script will be moved into the install-identity script
```
ctediz: Will need a way to specify which tenant should also go into the Identity web.config

Fabric.Identity.API/scripts/registration.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
ctediz: Test config that got in, will be removed

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -14,6 +14,7 @@ param(
     [switch] $quiet
 )
 Import-Module -Name .\Install-Identity-Utilities.psm1 -Force
+Import-Module -Name .\Install-IdPSS.psm1 -Force
```
ctediz: Wrong script name, but will update when moving the Alter-Config script into here

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+        $tenantSettings.AppendChild($setting) | Out-Null
+    }
+    else{
+        $existingSetting.secret = $clientSecret
+        $existingSetting.clientId = $clientId
+    }
+    $installationConfig.Save("$installConfigPath") | Out-Null
+}
+
+function Set-AppSettings($appDirectory, $appSettings){
```
ctediz: A similar Set-EnvironmentVariable/Add-EnvironmentVariable exists fabric-install-utilities for these two, but they go into slightly different locations (appSettings vs environmentVariables) with different encapsulating elements. Not sure if this should also go into the fabric-install-utilities eventually.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+# TODO: Differentiate between idpss and identity application for extra/different permissions
+if($null -ne $tenants) {
+    foreach($tenant in $tenants) {
+        Write-Host "Enter credentials for specified tenant $tenant"
+        $credential = Get-Credential
+
+        Connect-AzureADTenant -tenantId $tenant -credential $credential
+
+        # TODO: This will potentially redirect to an application that may not be installed yet?
```
ctediz: Note ^

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -0,0 +1,47 @@
...
+        $app = New-FabricAzureADApplication -appName 'Identity Provider Search Service' -replyUrls $replyUrls
+        $clientId = $app.AppId
+        $clientSecret = Get-FabricAzureADSecret -objectId $app.ObjectId
+
+        Disconnect-AzureAD
+        Add-InstallationTenantSettings -configSection "common" -tenantId $tenant -clientSecret $clientSecret -clientId $clientId -installConfigPath $installConfigPath
+
+        # Manual process, need to give consent this way for now
+        Start-Process -FilePath  "https://login.microsoftonline.com/$tenant/oauth2/authorize?client_id=$clientId&response_type=code&state=12345&prompt=admin_consent"
```
ctediz: If no redirect url is set, above, then this gives an error about not having a redirect url and fails the admin consent.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+        Connect-AzureADTenant -tenantId $tenant -credential $credential
+
+        # TODO: This will potentially redirect to an application that may not be installed yet?
+        $app = New-FabricAzureADApplication -appName 'Identity Provider Search Service' -replyUrls $replyUrls
+        $clientId = $app.AppId
+        $clientSecret = Get-FabricAzureADSecret -objectId $app.ObjectId
+
+        Disconnect-AzureAD
+        Add-InstallationTenantSettings -configSection "common" -tenantId $tenant -clientSecret $clientSecret -clientId $clientId -installConfigPath $installConfigPath
```
ctediz: Putting into common directly, should I make a IdPSS config section to go along or is this fine?

Fabric.Identity.API/scripts/Alter-Config.ps1
```diff
@@ -0,0 +1,47 @@
...
+$fabricInstallUtilities = ".\Fabric-Install-Utilities.psm1"
+if (!(Test-Path $fabricInstallUtilities -PathType Leaf)) {
+    Write-DosMessage -Level "Warning" -Message "Could not find fabric install utilities. Manually downloading and installing"
+    Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -Headers @{"Cache-Control" = "no-cache"} -OutFile $fabricInstallUtilities
+}
+#Import-Module -Name $fabricInstallUtilities -Force
+Import-Module -Name "$PSScriptRoot\Install-IdPSS-Utilities.psm1" -Force
+Import-Module -Name "$PSScriptRoot\Fabric-Install-Utilities.psm1" -Force
+
+#$installSettings = Get-InstallationSettings -configSection "common" -installConfigPath ".\install.config"
```
ctediz: Was playing with using the existing Fabric-install-utilities Get-InstallationSettings, but working directly with the xml seemed like the easiest quick choice for now.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+
+function New-FabricAzureADApplication() {
+    param(
+        [Parameter(Mandatory=$true)]
+        [string] $appName,
+        [Parameter(Mandatory=$true)]
+        [string[]] $replyUrls
+    )
+    # Get read permissions
+    $readResource = Get-GraphApiReadPermissions
```
ctediz: Identity install probably has extra/different permissions that will need to be set

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+        Import-Module AzureAD -Force
+    }
+}
+else {
+    Write-Host "Installing AzureAD at $($azureAD.FullName)"
+    Import-Module -Name $azureAD.FullName
+}
+
+function Get-GraphApiReadPermissions() {
+    $aad = (Get-AzureADServicePrincipal | Where-Object {$_.ServicePrincipalNames.Contains("https://graph.microsoft.com")})[0]
```
ctediz: This is kind of awkward, but didn't immediately see a good way to do this, can revisit

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+    }
+    else {
+        Set-AzureADApplication -ObjectId $app.ObjectId -RequiredResourceAccess $readResource -Oauth2AllowImplicitFlow $true -ReplyUrls $replyUrls
+    }
+
+    return $app
+}
+
+function Get-FabricAzureADSecret([string] $objectId) {
+    # TODO: Remove and recreate on every install?
```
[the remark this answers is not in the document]
michaelvidal24: I like the idea of recreating it on every install - it allows us to naturally rotate our secrets which is a nice benefit. It also allows us to always make sure we have the right secret during install.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+            return $true
+        })]  
+        [string] $installConfigPath = "$(Get-CurrentScriptDirectory)\install.config"
+    )
+
+    $installationConfig = [xml](Get-Content $installConfigPath)
+    $tenantScope = $installationConfig.installation.settings.scope | Where-Object {$_.name -eq $configSection}
+    $tenantSettings = $tenantScope.SelectSingleNode('tenants')
+
+    # Add a tenant section if not exists
```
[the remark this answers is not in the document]
michaelvidal24: I think this is fine, we are trying to deprecate Fabric-Install-Utilities anyway.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,235 @@
...
+    #[string] $encryptionCertificateThumbprint, `
+    #[string] $appInsightsInstrumentationKey, `
+    param(
+        [string] $appDirectory,
+        [object[]] $clientSettings,
+        [string] $useAzure = $false
+    )
+    $appSettings = @{}
+
+    # TODO: Implement encryption for secrets
```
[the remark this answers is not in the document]
michaelvidal24: Yeah - we'll have to add the decryption logic in IdPSS as well.

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -11,9 +11,17 @@ param(
...
     [switch] $noDiscoveryService, 
-    [switch] $quiet
+    [switch] $quiet,
+    [ValidateScript({
+        if (!(Test-Path $_)) {
+            throw "Path $_ does not exist. Please enter valid path to the IdentitySearchProvider web.config."
+        }
+        return $true
+    })]
+    [string] $idpssDirectoryPath
```
michaelvidal24: I think we should either get this path by interogating IIS (we do this in the register.ps1 script), or have it as an install.config variable. We want to stick to the standard paramters on the script.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -0,0 +1,47 @@
...
+            throw "Path $_ does not exist. Please enter valid path to the install.config."
+        }
+        if (!(Test-Path $_ -PathType Leaf)) {
+            throw "Path $_ is not a file. Please enter a valid path to the install.config."
+        }
+        return $true
+    })] 
+    [string] $installConfigPath = "$PSScriptRoot\install.config",
+    [Parameter(Mandatory=$true)]
+    [string[]] $tenants,
```
michaelvidal24: I think these should become install.config params, for the reasons stated above.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
michaelvidal24: I'm thinking we put all IdPSS config and Identity config into the "identity" config section and treat them as one logical application, i.e. they get installed with the same script, and their config lives in the same place in install.config. Of course we would need to spread that config across two web.configs, but to the end user I want the two apps to feel like one.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
michaelvidal24: We should validate that we have a proper redirect URI then, at least that its not null or empty...

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
ryanorbaker: do we need to write the secret to the config file if we re-create each time?

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    if (($null -eq $installed) -or ($installed.Version.CompareTo($minVersion) -lt 0)) {
+        Write-Host "Installing AzureAD from Powershell Gallery"
+        Install-Module AzureAD -Scope CurrentUser -MinimumVersion $minVersion -Force
+        Import-Module AzureAD -Force
+    }
+}
+else {
+    Write-Host "Installing AzureAD at $($azureAD.FullName)"
```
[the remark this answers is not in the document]
ctediz: Hmm, i'm not sure what happens if the module is installed locally and hits this section, can take a look at it

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
ctediz: Yes, this script registers only, so the install script will need to grab these values somewhere in order to move them into the proper web.configs.  The reason for splitting these two out, is that the registration can't really be automated at the moment since we need human interaction to give admin consent to an application

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -34,6 +35,13 @@ if(!(Test-IsRunAsAdministrator))
 Write-DosMessage -Level "Information" -Message "Using install.config: $installConfigPath"
 $installSettingsScope = "identity"
 $installSettings = Get-InstallationSettings $installSettingsScope -installConfigPath $installConfigPath
+
+$idpssConfig = $installSettings.identityProviderSearchServiceConfig
+if($null -eq $idpssConfig) {
+    $idpssDirectoryPath = Get-WebConfigPath -service "IdentityProviderSearchService" -discoveryServiceUrl $installSettings.discoveryService -noDiscoveryService $noDiscoveryService -quiet $quiet
+    Add-InstallationSetting -configSection $installSettingsScope -configSetting "identityProviderSearchServiceConfig" -configValue $idpssDirectoryPath -installConfigPath $installConfigPath | Out-Null
```
michaelvidal24: Is there a reason we are writing this value into the install.config instead of just setting it as a variable for later use?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -62,6 +70,12 @@ if(!$noDiscoveryService){
     Register-IdentityWithDiscovery -iisUserName $iisUser.UserName -metadataConnStr $metadataDatabase.DbConnectionString -version $installApplication.version -identityServerUrl $identityServiceUrl
 }
 
+$useAzure = $installSettings.useAzure
+if($null -eq $useAzure) {
+    $useAzure = $false
+    Add-InstallationSetting -configSection $installSettingsScope -configSetting "useAzure" -configValue "$useAzure" -installConfigPath $installConfigPath | Out-Null
```
michaelvidal24: Same question as above...

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -632,7 +632,61 @@ function Get-DefaultApplicationEndpoint([string] $appName, [string] $appEndPoint
 }
 
 function Get-FullyQualifiedMachineName() {
-	return "https://$env:computername.$((Get-WmiObject Win32_ComputerSystem).Domain.tolower())"
+    return "https://$env:computername.$((Get-WmiObject Win32_ComputerSystem).Domain.tolower())"
+}
+
+function Get-ApplicationUrl($serviceName, $discoveryServiceUrl){
```
michaelvidal24: Does Dos-InstallUtilities have functions for this?

Fabric.Identity.API/scripts/Install-Identity.ps1
ctediz: I don't think this one needs to be written to the config, so it can go

Fabric.Identity.API/scripts/Install-Identity.ps1
ctediz: This one I feel should stay, if not here, somewhere else, since this will be configurable, it would be easier to toggle it going forward by setting the value to true/false instead of writing the whole element manually

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: It does not

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -62,6 +69,18 @@ if(!$noDiscoveryService){
...
 
+$useAzure = $installSettings.useAzureAD
+if($null -eq $useAzure) {
+    $useAzure = $false
+    Add-InstallationSetting -configSection $installSettingsScope -configSetting "useAzureAD" -configValue "$useAzure" -installConfigPath $installConfigPath | Out-Null
+}
+
+$useWindows = $installSettings.useWindowsAD
+if($null -eq $useWindows) {
+    $useWindows = $false
```
michaelvidal24: Default to true?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -34,6 +35,12 @@ if(!(Test-IsRunAsAdministrator))
 Write-DosMessage -Level "Information" -Message "Using install.config: $installConfigPath"
 $installSettingsScope = "identity"
 $installSettings = Get-InstallationSettings $installSettingsScope -installConfigPath $installConfigPath
+
+$idpssConfig = $installSettings.identityProviderSearchServiceConfig
+if($null -eq $idpssConfig) {
+    $idpssDirectoryPath = Get-WebConfigPath -service "IdentityProviderSearchService" -discoveryServiceUrl $installSettings.discoveryService -noDiscoveryService $noDiscoveryService -quiet $quiet
```
michaelvidal24: Don't try to get path from installConfig, just get from Discovery service and IIS.

ctediz:
Fixing comments in a different PR

PR #257, opened 2018-12-04
User/ken/identityinstall

hckenmiller:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/257)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -279,7 +279,7 @@ function Clear-IdentityEnvironmentAzureSettings {
     $settings = $content.configuration
     $environmentVariables = $settings.ChildNodes.aspNetCore.environmentVariables
 
-    $azureSettings = ($environmentVariables.ChildNodes | Where-Object {$_.name.StartsWith("AzureActiveDirectorySettings") -or $_.name -eq "AzureAuthenticationEnabled"})
+    $azureSettings = ($environmentVariables.ChildNodes | Where-Object {$_.name.StartsWith("AzureActiveDirectorySettings")})
     
     foreach($setting in $azureSettings) {
         Write-Host "Cleaning up setting: $($setting.name)"
@@ -488,7 +488,7 @@ function Register-Identity {
         -setting $claimsIssuerText
 
    if($null -ne $claimsIssuer) {
-    Write-Host "Enter Credentials for identity claims issuer: $claimsIssuer"
+    Write-Host "Enter credentials for $appName specified tenant: $claimsIssuer"
 	Connect-AzureADTenant -tenantId $claimsIssuer
 
 	$app = New-FabricAzureADApplication -appName $appName -replyUrls $replyUrls
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -271,7 +271,7 @@ function Clear-IdentityAppSettings {
 
 }
 
-function Clear-IdentityEnvironmentSettings {
+function Clear-IdentityEnvironmentAzureSettings {
     param(
         [string] $environmentSettingPath
     )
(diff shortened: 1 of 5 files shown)


Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -262,6 +271,24 @@ function Clear-IdentityAppSettings {
 
 }
 
+function Clear-IdentityEnvironmentSettings {
```
michaelvidal24: Is this function appropriately named? Seems like we are only clearing the azure settings?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -57,6 +57,17 @@ if(!$noDiscoveryService){
 }
 $identityServiceUrl = Get-ApplicationEndpoint -appName $installSettings.appName -applicationEndpoint $installSettings.applicationEndPoint -installConfigPath $installConfigPath -scope $installSettingsScope -quiet $quiet
 
+$idpssConfig = $installSettings.identityProviderSearchServiceConfig
```
michaelvidal24: @ctediz - Do these changes conflict with some of yours?

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -37,9 +39,15 @@ if($null -ne $tenants) {
...
 
         # Manual process, need to give consent this way for now
         Start-Process -FilePath  "https://login.microsoftonline.com/$tenant/oauth2/authorize?client_id=$clientId&response_type=code&state=12345&prompt=admin_consent"
     }
-}
\ No newline at end of file
+}
+
+#identity registration
+Register-Identity -appName "Identity Service" -replyUrls $replyUrls -configSection $installSettingsScope -installConfigPath $installConfigPath
```
michaelvidal24: Should we add a command line parameter so the user can decide to register idpss or identity separately?

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -37,9 +39,15 @@ if($null -ne $tenants) {
             -tenantId $tenant `
             -clientSecret $clientSecret `
             -clientId $clientId `
-            -installConfigPath $installConfigPath
+            -installConfigPath $installConfigPath `
+			-appName $appName
```
ctediz: $appName doesn't look like it's being set

hckenmiller: I removed the appName variable by accident in the merge. Good catch. I will add it back.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: Yah, I remember unchecking it in the merge, yet here it is again duplicated. Thanks

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
michaelvidal24: The reason I ask is because we only need to register Identity in one of the Azure tenants. IdPSS needs to be registered in all tenants that are in the environment...so it would be nice to either allow only IdPSS or both IdPSS and Identity.

hckenmiller: Yah, currently I use the claimsissuer as the tenant used to register identity. I think prompting them for (claimsissuer and allowedtenants) during the identity registration, would make sense for scenario two, and no nodes are added for scenario one when Idpss is registered.

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -8,7 +8,8 @@ param(
         }
         return $true
     })] 
-    [string] $installConfigPath = "$PSScriptRoot\install.config"
+    [string] $installConfigPath = "$PSScriptRoot\install.config",
+	[string[]] $registerApps = ("Identity Service", "Identity Provider Search Service")
```
ctediz: Does it make sense to make these two into switches instead, similar to $quiet in install-identity

hckenmiller: sounds good

Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
hckenmiller: added switches to cover this, and removed all the node creation and hard coded the necessary install.config nodes, since they will need to manually enter all tenant info.

PR #259, opened 2018-12-04
Add Azure AD OID as Subject Id

hc-jordenlowe:
Add Azure AD OID as Subject Id

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/259)

### DIFF
--- a/Fabric.Identity.API/Quickstart/Account/AccountController.cs
+++ b/Fabric.Identity.API/Quickstart/Account/AccountController.cs
@@ -15,13 +15,12 @@ using System.Linq;
 using System.Security.Claims;
 using System.Security.Principal;
 using System.Threading.Tasks;
-using Fabric.Identity.API;
 using Fabric.Identity.API.Configuration;
 using Fabric.Identity.API.Events;
 using Fabric.Identity.API.Management;
 using Fabric.Identity.API.Persistence;
 using Fabric.Identity.API.Services;
-using Microsoft.AspNetCore.Authentication;
+using IFabricClaimsService = Fabric.Identity.API.Services.IClaimsService;
 using IdentityServer4.Events;
 using IdentityServer4.Extensions;
 using Serilog;
@@ -32,6 +31,7 @@ namespace IdentityServer4.Quickstart.UI
 
     using Fabric.Identity.API.Exceptions;
     using Fabric.Identity.API.Models;
+    using IdentityServer4.Models;
 
     /// <summary>
     /// This sample controller implements a typical login/logout/provision workflow for local and external accounts.
@@ -47,6 +47,7 @@ namespace IdentityServer4.Quickstart.UI
         private readonly IAppConfiguration _appConfiguration;
         private readonly ILogger _logger;
         private readonly IExternalIdentityProviderService _externalIdentityProviderService;
+        private readonly IFabricClaimsService _claimsService;
         private readonly AccountService _accountService;
         private readonly UserLoginManager _userLoginManager;
 
@@ -61,6 +62,7 @@ namespace IdentityServer4.Quickstart.UI
             IUserStore userStore,
             ILogger logger,
             IExternalIdentityProviderService externalIdentityProviderService,
+            IFabricClaimsService claimsService,
             AccountService accountService,
             GroupFilterService groupFilterService,
             TestUserStore users = null)
@@ -73,6 +75,7 @@ namespace IdentityServer4.Quickstart.UI
             _logger = logger;
             _externalIdentityProviderService = externalIdentityProviderService;
             _accountService = accountService;
+            _claimsService = claimsService;
             _groupFilterService = groupFilterService;
             _userLoginManager = new UserLoginManager(userStore, _logger);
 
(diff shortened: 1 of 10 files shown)


Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+                ClientId = context?.ClientId,
+                UserId = userIdClaim.Value,
+                Provider = provider
+            };
+
+            if (this.IsAzureEnabled(schemaItem))
+            {
+                if (this.ValidateAzureAD(claims))
+                {
+                    var issuerClaim = claims.FirstOrDefault(x => x.Type == JwtClaimTypes.Issuer);
```
michaelvidal24: Why would we throw an exception if the issuer claim is valid and present? I think we might just be missing a `!` operator on this check?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (userIdClaim == null)
+            {
+                throw new MissingUserClaimException(ExceptionMessageResources.MissingUserClaimMessage);
+            }
+
+            return userIdClaim;
+        }
+
+        private bool IsAzureEnabled(string schemaItem) =>
```
michaelvidal24: This is really more than merely AzureAD being enabled, its that AzureAd is enabled, and the authentication scheme is azure for the current user - perhaps we should rename this one.

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+            var result = new ClaimsResult()
+            {
+                ClientId = context?.ClientId,
+                UserId = userIdClaim.Value,
+                Provider = provider
+            };
+
+            if (this.IsAzureEnabled(schemaItem))
+            {
+                if (this.ValidateAzureAD(claims))
```
michaelvidal24: Are we missing a `!` operator here? I think we only want to throw this exception if we don't have a valid issuer, not if we do.

Fabric.Identity.API/Services/ClaimsService.cs
michaelvidal24: Should we rename this function? It's really IsAzureADEnabledAndAuthSchemeisAzureAD.

Fabric.Identity.API/Quickstart/Account/ClaimsResult.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
ryanorbaker: is there any reason this is not in models?
looks like it was copied in from another project / base?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.AspNetCore.Http.Authentication;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Security.Claims;
+
+namespace Fabric.Identity.API.Services
+{
+    public class ClaimsService
```
ryanorbaker: would this be easier for testing if it had an interface?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+    {
+        private readonly IAppConfiguration _appConfiguration;
+
+        public ClaimsService(
+            IAppConfiguration appConfiguration)
+        {
+            _appConfiguration = appConfiguration;
+        }
+
+        public ClaimsResult GenerateClaimsForIdentity(List<Claim> claims, AuthenticateInfo info, AuthorizationRequest context, string schemaItem)
```
ryanorbaker: if you're passing in info, can't you get schemaItem from that?

Fabric.Identity.API/Quickstart/Account/AccountController.cs
```diff
@@ -226,87 +232,52 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl)
...
-            {
-                userIdClaim = claims.FirstOrDefault(x => x.Type == ClaimTypes.NameIdentifier);
-            }
-
-            if (userIdClaim == null)
-            {
-                throw new MissingUserClaimException(ExceptionMessageResources.MissingUserClaimMessage);
-            }
-
             var schemaItem = info.Properties.Items.FirstOrDefault(i => i.Key == "scheme");
```
ryanorbaker: I know it's existing code, but would it be Schem**_e_** Item?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private bool ValidateAzureAD(List<Claim> claims)
+        {
+            var issuerClaim = claims.FirstOrDefault(x => x.Type == JwtClaimTypes.Issuer);
+            if (issuerClaim == null)
+            {
+                throw new MissingIssuerClaimException(ExceptionMessageResources.MissingIssuerClaimMessage);
+            }
+
+            return !this._appConfiguration.AzureActiveDirectorySettings.IssuerWhiteList.Contains(issuerClaim.Issuer);
```
ryanorbaker: so the validate function returns true if the issuer is NOT in the whitelist?    
Would it be more clear to call the function IsAzureADInvalid ? 

or return true if it's in the whitelist and change the logic that uses it.

Fabric.Identity.API/Services/ClaimsService.cs
ryanorbaker: ISAzureEnabledForScheme ?

hc-jordenlowe: This function checks to see if the external token is AzureAD, so im going to go with IsExternalTokenAzureAD

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: yeah the ! needs to be moved.

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -0,0 +1,147 @@
...
+            claims.Remove(userIdClaim);
+
+            //get the client id from the auth context
+            result.AdditionalClaims = this.GenerateAdditionalClaims(claims);
+            result.AuthenticationProperties = this.GenerateAuthenticationProperties(info);
+
+            return result;
+        }
+
+        public string GetSubjectId(List<Claim> claims, User user, string schemaItem)
```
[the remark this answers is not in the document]
hc-jordenlowe: ok

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: it was a refactor that moved the ! inside the function, ill just bring it back out.

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: So if you get into here, the AzureAD issuer claim is not valid, Based on the function above.  So we are getting the issuer that is being used to help diagnose the issue.

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: thought it would be messy to get it twice considering it requires doing this:  info.Properties.Items.FirstOrDefault(i => i.Key == "scheme").value

By just passing it into the function, the function doesnt need to know about how to get this.  Also reduces complexity.  Say if we have to refactor in the future, there is one spot and not two that must be modified.

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: The controller is not the easiest to test, but this will help with that.  I can make that change.

Fabric.Identity.API/Quickstart/Account/ClaimsResult.cs
hc-jordenlowe: The refactoring started in that namespace.

While I like the Models folder for this, I feel the intent for Models folder were Models that were external facing types.  I dont  see this being used externally, I dont see alternatives to where this could go.

Fabric.Identity.API/Quickstart/Account/AccountController.cs
hc-jordenlowe: probably

Fabric.Identity.API/Services/ClaimsService.cs
michaelvidal24: That makes more sense, I like it!

Fabric.Identity.API/Services/ClaimsService.cs
michaelvidal24: I get it now, I missed the `!` inside the function.

Fabric.Identity.API/Services/ClaimsService.cs
hc-jordenlowe: Ok, i think i got it.  I moved more of the initialization into the function and everything will be generated and returned in the ClaimsResult.  I think this makes the AccountController function cleaner and hides some of the magic of parsing everything.

PR #260, opened 2018-12-06
Azure settings only written when $useAzure -eq $true

hckenmiller:
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/260)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -271,24 +271,6 @@ function Clear-IdentityProviderSearchServiceWebConfigAzureSettings {
 
 }
 
-function Clear-IdentityEnvironmentAzureSettings {
-    param(
-        [string] $environmentSettingPath
-    )
-    $content = [xml](Get-Content $environmentSettingPath)
-    $settings = $content.configuration
-    $environmentVariables = $settings.ChildNodes.aspNetCore.environmentVariables
-
-    $azureSettings = ($environmentVariables.ChildNodes | Where-Object {$_.name.StartsWith("AzureActiveDirectorySettings") -or $_.name -eq ("WindowsAuthenticationEnabled") -or $_.name -eq ("AzureAuthenticationEnabled")})
-    
-    foreach($setting in $azureSettings) {
-        Write-Host "Cleaning up setting: $($setting.name)"
-        $environmentVariables.RemoveChild($setting) | Out-Null
-    }
-    $content.Save("$environmentSettingPath")
-
-}
-
 function Set-IdentityProviderSearchServiceWebConfigSettings {
     param(
         [string] $encryptionCertificateThumbprint,
@@ -503,7 +485,6 @@ function Set-IdentityEnvironmentAzureVariables {
         [string] $useWindows = $true,
         [System.Security.Cryptography.X509Certificates.X509Certificate2] $encryptionCert
     )
-	Clear-IdentityEnvironmentAzureSettings -environmentSettingPath $appConfig\web.config
     $environmentVariables = @{}
 
 	if($useAzure -eq $true)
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -300,52 +300,55 @@ function Set-IdentityProviderSearchServiceWebConfigSettings {
         [System.Security.Cryptography.X509Certificates.X509Certificate2] $encryptionCert,
         [string] $appName
     )
-
-    # Alter IdPSS web.config for azure
-    $clientSettings = @()
-    $clientSettings += Get-ClientSettingsFromInstallConfig -installConfigPath $installConfigPath -appName $appName
-
-    Clear-IdentityProviderSearchServiceWebConfigAzureSettings -webConfigPath $webConfigPath
+	Clear-IdentityProviderSearchServiceWebConfigAzureSettings -webConfigPath $webConfigPath
```
michaelvidal24: @hckenmiller , @ctediz - what is the purpose of clearing the settings first? Asking because during a deployment, the web.config is overwritten in the upgrade scenario, so we are always starting from a known state.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -500,50 +503,53 @@ function Set-IdentityEnvironmentAzureVariables {
...
-    $clientSettings = Get-ClientSettingsFromInstallConfig -installConfigPath $installConfigPath -appName "Identity Service"
-    $allowedTenants += Get-SettingsFromInstallConfig -installConfigPath $installConfigPath `
-        -scope $scope `
-        -setting "allowedTenants"
-
-    $claimsIssuer += Get-SettingsFromInstallConfig -installConfigPath $installConfigPath `
-        -scope $scope `
-        -setting "claimsIssuerTenant"
-
 	Clear-IdentityEnvironmentAzureSettings -environmentSettingPath $appConfig\web.config
```
michaelvidal24: Same question as the other comment.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
hckenmiller: Running install-identity, shouldn't the azuresettings getting cleared out if they changed from $useAzure = $true to $false in the install.config?

michaelvidal24: I guess what I am trying to communicate is that every time we run an installation, we blow away the entire web.config and it gets re deployed with our default settings, so we start from a known state, which doesn't include any Azure settings. So I am not sure there is anything to clear...

hckenmiller: Ok. I get it. I was just basing it upon running install-identity. If that is the case then I could probably remove it.

ctediz: I think I know what happened, it clears out for identity but not idpss since this script doesn't install idpss, so the web.config wasn't in a clean state when updating the idpss web.config

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -500,50 +503,53 @@ function Set-IdentityEnvironmentAzureVariables {
...
+		$allowedTenants += Get-SettingsFromInstallConfig -installConfigPath $installConfigPath `
+			-scope $scope `
+			-setting "allowedTenants"
+
+		$claimsIssuer += Get-SettingsFromInstallConfig -installConfigPath $installConfigPath `
+			-scope $scope `
+			-setting "claimsIssuerTenant"
+
+		# Set Azure Settings
+		$environmentVariables.Add("AzureActiveDirectorySettings_Authority", "https://login.microsoftonline.com/common")
```
michaelvidal24: The separator for these requires a double underscore, for example: `AzureActiveDirectorySettings_Authority` or `AzureActiveDirectorySettings_Scope_0`.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
hckenmiller: Found issue running install-identity locally, not updating web.config, because of a bad Fabric.Identity.API.zip file. Jason's IdPSS only gets cleared out if you run the DosInstaller IdPSS step, so will leave the clear function for his.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
hckenmiller: Good catch. Thanks.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -500,50 +503,53 @@ function Set-IdentityEnvironmentAzureVariables {
...
+				$environmentVariables.Add("AzureActiveDirectorySettings_ClientSecret", $encryptedSecret)
+			}
+			else{
+				$environmentVariables.Add("AzureActiveDirectorySettings_ClientSecret", $secret)
+			}
+
+		foreach($allowedTenant in $allowedTenants)
+		{
+		  $index = $allowedTenants.IndexOf($allowedTenant)
+		  $environmentVariables.Add("AzureActiveDirectorySettings_IssuerWhiteList_$index", "https://sts.windows.net/" + $allowedTenant)
```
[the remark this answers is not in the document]
hckenmiller: Good catch. Thanks.

PR #263, opened 2018-12-07
User/jason/azureadpsscriptchanges

ctediz:
Make Install-Identity not dependent on install-idpss-utilities to not pull down AzureAD module when not needed.


Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -5,10 +5,8 @@ Import-Module -Name $fabricInstallUtilities -Force
 $minVersion = [System.Version]::new(2, 0, 2 , 4)
 $azureAD = Get-Childitem -Path ./**/AzureAD.psm1 -Recurse
 if ($azureAD.length -eq 0) {
-    $installed = Get-Module -Name AzureAD
-    if ($null -eq $installed) {
-        $installed = Get-InstalledModule -Name AzureAD
-    }
+    # Do not show error when AzureAD is not installed, will install instead
+    $installed = Get-InstalledModule -Name AzureAD -ErrorAction "silentlycontinue"
```
ctediz: I thought 'silentlycontinue' made more sense than setting erroraction to 'stop' instead and wrapping in a try catch, the end result is the same and $installed stays null if not found

PR #264, opened 2018-12-07
User/jason/moreidpss

ctediz:
Add retry logic for setting/creating azure secrets
Give Idpss App Pool users permissions to read private key cert

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/264)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -20,8 +20,20 @@ else {
 }
 
 function Get-GraphApiReadPermissions() {
-    $aad = (Get-AzureADServicePrincipal | Where-Object {$_.ServicePrincipalNames.Contains("https://graph.microsoft.com")})[0]
-    $directoryRead = $aad.AppRoles | Where-Object {$_.Value -eq "Directory.Read.All"}
+    try {
+        $aad = (Get-AzureADServicePrincipal | Where-Object {$_.ServicePrincipalNames.Contains("https://graph.microsoft.com")})[0]
+    }
+    catch {
+        Write-DosMessage -Level "Error" -Message "Was not able to get the Microsoft Graph API service principal."
+        throw
+    }
+
+    $directoryReadName = "Directory.Read.All"
+    $directoryRead = $aad.AppRoles | Where-Object {$_.Value -eq $directoryReadName}
+    if($null -eq $directoryRead) {
+        Write-DosMessage -Level "Error" -Message "Was not able to find permissions $directoryReadName."
+        throw
+    }
 
     # Convert to proper resource...
     $readAccess = [Microsoft.Open.AzureAD.Model.RequiredResourceAccess]@{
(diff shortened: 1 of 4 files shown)


Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -35,7 +35,9 @@ Write-DosMessage -Level "Information" -Message "Using install.config: $installCo
 $installSettingsScope = "identity"
 $installSettings = Get-InstallationSettings $installSettingsScope -installConfigPath $installConfigPath
 
-$idpssConfig = Get-WebConfigPath -service "IdentityProviderSearchService" -discoveryServiceUrl $installSettings.discoveryService -noDiscoveryService $noDiscoveryService -quiet $quiet
+$idpssName = "IdentityProviderSearchService"
```
ctediz: This is assuming IdentityProviderSearchService app name never changes, is that a valid assumption?

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -64,9 +76,22 @@ function Remove-AzureADClientSecret{
...
+        do {
+            Write-Host "Removing existing password credential named `"$($encoding.GetString($key.CustomKeyIdentifier))`" with id $($key.KeyId)"
+            try {
+                Remove-AzureADApplicationPasswordCredential -ObjectId $objectId -KeyId $key.KeyId -ErrorAction 'stop'
+                $completed = $true
+            }
+            catch {
+                Write-DosMessage -Level "Warning" -Message "An error occurred trying to remove the Azure application secret named `"$($encoding.GetString($key.CustomKeyIdentifier))`" with id $($key.KeyId). Retrying.."
+                Start-Sleep 3
+                $retryCount++
```
michaelvidal24: Unless I missed something, it doesn't look like we break out of the do while if we exceed n number of retries...

Fabric.Identity.API/scripts/Install-Identity.ps1
michaelvidal24: The name registered with Discovery should be safe to assume it won't change. We can get the app pool details by mapping from the discoveryservice url to IIS config, rather than assuming the app pool name matches the app name....

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
ctediz: In the try catch line 87 if no error is thrown it continues

michaelvidal24: right, but what if it keeps throwing an exception? I don't see where we short circuit the do while loop in $completed never gets flipped to $true...

ctediz: Oh I see, ya i missed that check

Fabric.Identity.API/scripts/Install-Identity.ps1
ctediz: Not sure I follow your statement about app pool names matching

PR #265, opened 2018-12-10
UserId needed to be subject id for Azure AD

hc-jordenlowe:
UserId needed to be subject id for Azure AD

Moved around some private functions to keep all public functions together.

Added new function "GetEffectiveUserId".  Based on conversation with @michaelvidal24 , The user Id that is used to store a user in the database was being set to values for AD, even though the user was logging into Identity with an Azure AD account.  This will select the subject ID if it is from Azure AD, else it will use the user Id claim that was pulled in the "GetUserIdClaim" step.

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/265)

### DIFF
--- a/Fabric.Identity.API/Services/ClaimsService.cs
+++ b/Fabric.Identity.API/Services/ClaimsService.cs
@@ -24,6 +24,12 @@ namespace Fabric.Identity.API.Services
             _appConfiguration = appConfiguration;
         }
 
+        /// <summary>
+        /// Generates all information needed to create an access token for a given user.
+        /// </summary>
+        /// <param name="info">Authentication information</param>
+        /// <param name="context">Authorization Request</param>
+        /// <returns>Returns all the information into an object call ClaimsResult</returns>
         public ClaimsResult GenerateClaimsForIdentity(AuthenticateInfo info, AuthorizationRequest context)
         {
             CheckWhetherArgumentIsNull(info, nameof(info));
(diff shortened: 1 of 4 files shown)


Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -48,6 +54,62 @@ public ClaimsResult GenerateClaimsForIdentity(AuthenticateInfo info, Authorizati
...
+        /// <param name="claimInformation">Claims information from GenerateClaimsForIdentity</param>
+        /// <param name="user">User generated from logging into Identity</param>
+        /// <returns>The subject id</returns>
+        public string GetEffectiveSubjectId(ClaimsResult claimInformation, User user)
+        {
+            CheckWhetherArgumentIsNull(user, nameof(user));
+            CheckWhetherArgumentIsNull(claimInformation, nameof(claimInformation));
+
+            string subjectId = null;
+            if (this.IsExternalTokenAzureAD(claimInformation.SchemeItem))
```
bsmithvt: Not necessary now, but we may want to factor the `GetEffectiveSubjectId` and `GetEffectiveUserId` methods out into interfaces, and then encapsulate the AD and Azure AD specific logic in their own classes. Then as we add new IdPs, this class doesn't have to change. It's just a matter of adding a new IdP-specific class for extracting this type of information.

hc-jordenlowe: I was thinking about that as I put these new logic in, these if blocks are going to be a pain if we dont encapsulate them.  Perhaps a plugin model.  But totally agree.

PR #266, opened 2018-12-10
Only give idpss user private cert read permissions if azure is enabled

ctediz:
Small pr
Idpss app pool account only needs read permissions if azure is enabled as it only decrypts azure secrets.

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity.ps1
@@ -35,9 +35,20 @@ Write-DosMessage -Level "Information" -Message "Using install.config: $installCo
 $installSettingsScope = "identity"
 $installSettings = Get-InstallationSettings $installSettingsScope -installConfigPath $installConfigPath
 
+$useAzure = $installSettings.useAzureAD
+if($null -eq $useAzure) {
+    $useAzure = $false
+    Add-InstallationSetting -configSection $installSettingsScope -configSetting "useAzureAD" -configValue "$useAzure" -installConfigPath $installConfigPath | Out-Null
+}
+
+$useWindows = $installSettings.useWindowsAD
+if($null -eq $useWindows) {
+    $useWindows = $true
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1040,6 +1040,10 @@ function Find-IISAppPoolUser {
     }
 
     $username = (Get-Item (Join-Path 'IIS:\AppPools\' $appPool)).processModel.username
+
+    if($null -eq $username -or [string]::IsNullOrEmpty($username)) {
+        Write-DosMessage -Level "Error" -Message "Could not find user for application `"$applicationName`" with application pool `"$appPool`". Please verify that the application pool user is not a virtual account and the application pool has a valid user."
+    }
```
ctediz: Can't seem to find a way to even check if the app pool is set up with a virtual account.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1040,6 +1040,10 @@ function Find-IISAppPoolUser {
     }
 
     $username = (Get-Item (Join-Path 'IIS:\AppPools\' $appPool)).processModel.username
+
+    if($null -eq $username -or [string]::IsNullOrEmpty($username)) {
+        Write-DosMessage -Level "Error" -Message "Could not find user for application `"$applicationName`" with application pool `"$appPool`". Please verify that the application pool user is not a virtual account and the application pool has a valid user."
```
michaelvidal24: Should we make the `-Level "Fatal"` so that we ensure it throws an exception and causes the script to stop?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: Nice, that's exactly what was needed

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: Yup, didn't realize error continued on errorpreference was set to continue.

PR #271, opened 2018-12-12
User/jason/identitypermissions

ctediz:
Gives proper permissions app permissions to identity when registering with Azure

---
This change is [Reviewable](https://reviewable.io/reviews/healthcatalyst/fabric.identity/271)


ctediz:
Getting an invalid secret from azure, I believe it's related to bug #158975, will need to investigate

PR #274, opened 2018-12-20
wrapped Invoke-Sql calls in try/catch blocks and re-throw exception

bsmithvt:


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -303,8 +305,14 @@ function Get-IdentityDatabaseConnectionString([string] $identityDbName, [string]
...
     $identityDbConnStr = "Server=$($sqlServerAddress);Database=$($identityDbName);Trusted_Connection=True;MultipleActiveResultSets=True;"
 
-    Invoke-Sql $identityDbConnStr "SELECT TOP 1 ClientId FROM Clients" | Out-Null
-    Write-DosMessage -Level "Information" -Message "Identity DB Connection string: $identityDbConnStr verified"
+    try {
+        Invoke-Sql $identityDbConnStr "SELECT TOP 1 ClientId FROM Clients" | Out-Null
+        Write-DosMessage -Level "Information" -Message "Identity DB Connection string: $identityDbConnStr verified"
+    }
+    catch {
+        Write-Error $_.Exception -ErrorAction Stop
```
michaelvidal24: Why can't we use Write-DosMessage -Level "Fatal" here?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -318,8 +326,14 @@ function Get-MetadataDatabaseConnectionString([string] $metadataDbName, [string]
...
     $metadataConnStr = "Server=$($sqlServerAddress);Database=$($metadataDbName);Trusted_Connection=True;MultipleActiveResultSets=True;"
 
-    Invoke-Sql $metadataConnStr "SELECT TOP 1 RoleID FROM CatalystAdmin.RoleBASE" | Out-Null
-    Write-DosMessage -Level "Information" -Message "Metadata DB Connection string: $metadataConnStr verified"
+    try {
+        Invoke-Sql $metadataConnStr "SELECT TOP 1 RoleID FROM CatalystAdmin.RoleBASE" | Out-Null
+        Write-DosMessage -Level "Information" -Message "Metadata DB Connection string: $metadataConnStr verified"
+    }
+    catch {
+        Write-Error $_.Exception -ErrorAction Stop
```
michaelvidal24: Same comment as above...

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -26,6 +26,8 @@ else {
     Import-Module -Name $dosInstallUtilities.FullName
 }
 
+$ErrorActionPreference = "Stop"
```
michaelvidal24: By setting this, the default action on any error is going to be stop...which I think is largely fine...

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -548,7 +562,13 @@ function Add-DatabaseLogin([string] $userName, [string] $connString)
                 set @sql = 'CREATE LOGIN ' + QUOTENAME('$userName') + ' FROM WINDOWS'
                 EXEC sp_executesql @sql
             END"
-    Invoke-Sql $connString $query @{userName=$userName} | Out-Null
+    try {
+        Invoke-Sql $connString $query @{userName=$userName} | Out-Null
+    }
+    catch {
+        Write-Error $_.Exception -ErrorAction Continue
```
michaelvidal24: Is there a reason we are stopping for some and continuing for other?

bsmithvt: The original `Install-Identity-Windows` script had try/catch blocks for some `Invoke-Sql` calls while omitted it on others. Do we always want to halt if we can't do any SQL-related activity, including, for example, migrating data?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
bsmithvt: We can change this. We are using `Write-Error` in some places. I'll change to use the Dos library function.

michaelvidal24: We use Write-Error in Fabric-Install-Utilities, we shouldn't be using it anywhere else, we want to move to DosInstallUtilities for new stuff.

PR #278, opened 2019-01-28
User/bsmithvt/idpss unavailable bug

bsmithvt:


Fabric.Identity.API/Infrastructure/PolicyProvider.cs
```diff
@@ -11,7 +13,17 @@ public class PolicyProvider
         public readonly CircuitBreakerPolicy LdapErrorPolicy = Policy.Handle<LdapException>()
             .CircuitBreaker(5, TimeSpan.FromMinutes(5));
 
+        public readonly RetryPolicy IdPSearchServiceRetryPolicy = Policy.Handle<HttpRequestException>()
+            .RetryAsync(3);
+
         public readonly CircuitBreakerPolicy IdPSearchServiceErrorPolicy = Policy.Handle<HttpRequestException>()
-            .CircuitBreaker(5, TimeSpan.FromMinutes(5));
+            .CircuitBreakerAsync(3, TimeSpan.FromMinutes(1));
```
michaelvidal24: Is there a reason we want to close the circuit after one minute instead of 5?

bsmithvt: I figured 5 minutes was a rather long time to leave the circuit open, especially based on how this bug surfaces, but I can switch it back.

bsmithvt:
@michaelvidal24 I'm seeing almost immediate failure when testing with IdP SS unavailable, but I can set an explicit timeout.

michaelvidal24:
@bsmithvt - I think the service unavailable is one scenario. But another is if the service is stuck trying to stat up and can't respond. The http request will hang waiting for a response, so I think we should try to handle that with a smallish timeout on the http request. Ideally, that timeout would only apply for these specific requests, and not set a shared http client timeout.

bsmithvt:
@michaelvidal24 ok, good call. I wasn't aware that behavior was different (stopped vs. starting). I've updated the PR to specify an explicit timeout and handle exceptions associated with timeouts.

PR #286, opened 2019-02-19
refactored usage of DiscoveryServiceClient for IdPSS lookup to fix st…

bsmithvt:
…artup error

### DIFF
--- a/Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
+++ b/Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
@@ -25,18 +25,44 @@ namespace Fabric.Identity.API.Services
         private readonly PolicyProvider _policyProvider;
         private readonly HttpClient _httpClient;
         private readonly IHttpRequestMessageFactory _httpRequestMessageFactory;
+        private static DiscoveryServiceClient _discoveryServiceClient;
 
-        public IdPSearchServiceProvider(IAppConfiguration appConfig, ILogger logger, PolicyProvider policyProvider, HttpClient httpClient, IHttpRequestMessageFactory httpRequestMessageFactory)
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -25,18 +25,44 @@ public class IdPSearchServiceProvider : IExternalIdentityProviderService
...
+        }
+
+        private static string FormatUrl(string url)
+        {
+            return !url.EndsWith("/") ? $"{url}/" : url;
         }
 
         public async Task<ExternalUser> FindUserBySubjectId(string subjectId)
         {
+            if (_appConfig.UseDiscoveryService && string.IsNullOrWhiteSpace(_idPSSBaseUrl))
```
michaelvidal24: This looks like it will always be null, i.e. it doesn't seem we ever set it.

Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -25,18 +25,44 @@ public class IdPSearchServiceProvider : IExternalIdentityProviderService
...
+            {
+                var serviceRegistration =
+                    await _discoveryServiceClient.GetServiceAsync("IdentityProviderSearchService", 1);
+
+                if (!string.IsNullOrEmpty(serviceRegistration?.ServiceUrl))
+                {
+                    _appConfig.IdentityProviderSearchSettings.BaseUrl =
+                        FormatUrl(serviceRegistration.ServiceUrl);
```
michaelvidal24: I think FormatUrl and .EnsureTrailingSlash do the same thing, I think we can get rid of FormatUrl and just use EnsureTrailingSlash...

PR #290, opened 2019-02-26
Fixed an issue where we were adding two name claims to the claims col…

michaelvidal24:
…lection. This was breaking deserialization in the .net core openid connection auth middleware as it was expecting a string not an array.

Still need to beef up the unit tests around the changes.

### DIFF
--- a/Fabric.Identity.API/Management/UserLoginManager.cs
+++ b/Fabric.Identity.API/Management/UserLoginManager.cs
@@ -64,8 +64,15 @@ namespace Fabric.Identity.API.Management
 
         private string GetUserName(User user)
         {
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/Management/UserLoginManager.cs
```diff
@@ -65,6 +62,22 @@ private void SetNamePropertiesFromClaims(User user)
             }
         }
 
+        private string GetUserName(User user)
+        {
+            var userName = string.Empty;
+            var upn = user.Claims.FirstOrDefault(c => c.Type == FabricIdentityConstants.PublicClaimTypes.UserPrincipalName)?.Value;
+            var name = user.Claims.FirstOrDefault(c => c.Type == JwtClaimTypes.Name)?.Value;
```
bsmithvt: Minor optimization - could you change the `else if` below to an `else` and do this lookup for the `Name` claim in that?

PR #296, opened 2019-03-08
User/jason/idpsstenantlias

ctediz:
Changes for getting tenantAlias into the install.config and moving the registered applications (identity/idpss) into the registeredApplications section once the registration script runs


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -776,9 +776,7 @@ function Get-SettingsFromInstallConfig {
     $tempNode = $tenantScope.SelectSingleNode($setting)
     $settingList = @()
     foreach($nodeChild in $tempNode.variable){
-        if($nodeChild.name) {
-            $settingList += $nodeChild.name
-        }
+        $settingList += @{name = $nodeChild.name; alias = $nodeChild.alias}
```
miketabet: For future reference adding objects to arrays is less performant than using lists.  Since this is performance critical, probably not a huge deal and doesn't need to be fixed.

PR #297, opened 2019-03-08
Move tenant alias settings into identity/idpss web.configs

ctediz:
Move tenant alias settings into identity/idpss web.configs

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -739,7 +739,7 @@ function Get-ClientSettingsFromInstallConfig {
     $tenantScope = $installationConfig.installation.settings.scope | Where-Object {$_.name -eq "identity"}
     $tenants = $tenantScope.SelectSingleNode('registeredApplications')
 
-    $clientSettings = @()
+    $clientSettings = New-Object System.Collections.Generic.List[HashTable]
     foreach($tenant in $tenants.variable) {
       if ($tenant.appName -eq $appName)
       {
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -876,7 +877,8 @@ function Set-IdentityEnvironmentAzureVariables {
         # Set Azure Settings
         $environmentVariables.Add("AzureActiveDirectorySettings__Authority", "https://login.microsoftonline.com/common")
         $environmentVariables.Add("AzureActiveDirectorySettings__DisplayName", "Azure AD")
-        $environmentVariables.Add("AzureActiveDirectorySettings__ClaimsIssuer", "https://login.microsoftonline.com/" + $claimsIssuer)
+        $environmentVariables.Add("AzureActiveDirectorySettings__ClaimsIssuer", "https://login.microsoftonline.com/" + $claimsIssuer.name)
```
hc-jordenlowe: I noticed $claimsIssuer changed, but i dont see those changes, i take it the web service just parses differently.  Is that the case?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -901,7 +904,8 @@ function Set-IdentityEnvironmentAzureVariables {
         foreach($allowedTenant in $allowedTenants)
         {
           $index = $allowedTenants.IndexOf($allowedTenant)
-          $environmentVariables.Add("AzureActiveDirectorySettings__IssuerWhiteList__$index", "https://sts.windows.net/" + $allowedTenant + "/")
+          $environmentVariables.Add("AzureActiveDirectorySettings__IssuerWhiteList__$index", "https://sts.windows.net/" + $allowedTenant.name + "/")
```
hc-jordenlowe: Same idea with this, $allowedTenant, is this just parsed differently now?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: The way it's grabbed from the install.config is slightly different now.  It used to be a string, but was changed to be a hashtable to get the alias to come along with it as well.

PR #299, opened 2019-03-12
Fabric.Identity changes and Idpss installation script addition

hckenmiller:


Fabric.Identity.API/scripts/install.config
```diff
@@ -47,20 +50,27 @@
...
+      <!-- The Tenant identity is to be registered with -->
+      <claimsIssuerTenant>
+        <variable name="" alias=""/>
+      </claimsIssuerTenant>
+      <!-- The Tenants that are allowed through to the claims issuer  -->
+      <allowedTenants>
+        <variable name="" alias=""/>
+      </allowedTenants>
+    </scope>
+    <scope name="idpss">
```
ctediz: should avoid the acronym and have the full name typed out here

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1168,6 +1178,114 @@ function Set-IdentityUri {
...
+        [Parameter(Mandatory=$true)]
+        [Hashtable] $commonConfig,
+        [Parameter(Mandatory=$true)]
+        [Hashtable] $serviceConfig
+    )
+
+    Confirm-SettingIsNotNull -settingName "common.sqlServerAddress" -settingValue $commonConfig.sqlServerAddress
+    Confirm-SettingIsNotNull -settingName "common.metadataDbName" -settingValue $commonConfig.metadataDbName
+    Confirm-SettingIsNotNull -settingName "common.webServerDomain" -settingValue $commonConfig.webServerDomain
+    Confirm-SettingIsNotNull -settingName "common.clientEnvironment" -settingValue $commonConfig.clientEnvironment
```
ctediz: it doesn't look like we're using this, I would say to not include these possibly, but I'm not sure what else is dependent on the confirm-serviceconfig

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1168,6 +1178,114 @@ function Set-IdentityUri {
...
+                                @{
+                                    Name = "MetadataContext-Web.config Connection String";
+                                    Value = $metaDataConnectionString
+                                }
+                            )
+    
+    return $webDeployParameters
+}
+
+function Confirm-ServiceConfig{
```
ctediz: is this being used anywhere? not seeing it in this pr

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,67 @@ if ($fabricInstallerSecret){
...
+# Create log directory with read/write permissions for app pool user
+$idpssSettingsScope = "idpss"
+$idpssServiceUrl = Get-ApplicationEndpoint -appName $idpssConfigStore.appName -applicationEndpoint $idpssConfigStore.applicationEndPoint -installConfigPath $installConfigPath -scope $idpssSettingsScope -quiet $quiet
+    
+$idpssWebDeployParameters = Get-WebDeployParameters -serviceConfig $idpssConfigStore -commonConfig $commonConfigStore
+$idpssStandalonePath = ".\Fabric.IdentityProviderSearchService.zip"
+$idpssInstallerPath = "..\WebDeployPackages\Fabric.IdentityProviderSearchService.zip"
+$idpssInstallPackagePath = Get-WebDeployPackagePath -standalonePath $idpssStandalonePath -installerPath $idpssInstallerPath
+
+if($null -eq $credential)
```
ctediz: if credential isn't null, it's still trying to use the $idpssIISUser set inside the if, so it will be null down below in the publish portion right?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,67 @@ if ($fabricInstallerSecret){
...
+                      -AppPoolName $idpssConfigStore.appPoolName `
+                      -AppPoolCredential $idpssIISUser.credential `
+                      -AuthenticationType "Windows" `
+                      -WebDeploy
+
+$idpssAppPoolUser = $idpssIISUser.UserName
+
+# uses Discovery url, we should be able to remove since I Get-IISAppPoolUser above
+$idpssName = "IdentityProviderSearchService"
+#$idpssAppPoolUser = Find-IISAppPoolUser -applicationName $idpssName -discoveryServiceUrl $discoveryServiceUrl -noDiscoveryService $true -quiet $quiet
```
ctediz: this can probably be removed

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,67 @@ if ($fabricInstallerSecret){
...
+
+$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($selectedSite.physicalPath), $idpssName)
+New-AppRoot $idpssDirectory $idpssAppPoolUser
+
+Add-DatabaseSecurity $idpssAppPoolUser $installSettings.identityDatabaseRole $identityDatabase.DbConnectionString
+
+Register-ServiceWithDiscovery -iisUserName $idpssAppPoolUser -metadataConnStr $metadataDatabase.DbConnectionString -version $idpssInstallApplication.version -serverUrl $idpssServiceUrl `
+-serviceName $idpssName -friendlyName "Fabric.IdentityProviderSearchService" -description "The Fabric.IdentityProviderSearchService searches Identity Providers for matching users and groups.";
+
+#$registrationApiSecret = Add-RegistrationApiRegistration -identityServerUrl $identityServiceUrl -accessToken $accessToken
```
ctediz: left over? looks like it's on line 100

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: good catch. updating...

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: I initially used some functions, but changed methods/functionality in most cases. I will remove if they aren't being used anymore.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: I left a few comments in to remind myself what code was failing and why. I will cleanup before check in.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: I was waiting to decide whether registration was needed for idpss

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1168,6 +1178,114 @@ function Set-IdentityUri {
...
+    if(Test-Path $standalonePath){
+        return Resolve-Path $standalonePath
+    }
+    if(Test-Path $installerPath){
+        return Resolve-Path $installerPath
+    }
+    Write-DosMessage -Level "Fatal" -Message "Could not find the web deploy package at $standalonePath or $installerPath."
+}
+
+function Get-WebDeployParameters{
```
michaelvidal24: Should this be renamed to Get-IdpssWebDeployParameters?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -68,19 +75,13 @@ $installApplication = Publish-Application -site $selectedSite `
...
-
-
 Add-DatabaseSecurity $iisUser.UserName $installSettings.identityDatabaseRole $identityDatabase.DbConnectionString
 if(!$noDiscoveryService){
-    Register-IdentityWithDiscovery -iisUserName $iisUser.UserName -metadataConnStr $metadataDatabase.DbConnectionString -version $installApplication.version -identityServerUrl $identityServiceUrl
+    Register-ServiceWithDiscovery -iisUserName $iisUser.UserName -metadataConnStr $metadataDatabase.DbConnectionString -version $installApplication.version -serverUrl $identityServiceUrl `
+    -serviceName "IdentityService" -friendlyName "Fabric.Identity" -description "The Fabric.Identity service provides centralized authentication across the Fabric ecosystem.";
 }
 
+# uses discovery getting set to true causes identity service to be in error, should be fixed in more recent fabric.identity
```
michaelvidal24: Is this comment needed anymore?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -56,6 +56,13 @@ $appInsightsKey = Get-AppInsightsKey -appInsightsInstrumentationKey $installSett
 $sqlServerAddress = Get-SqlServerAddress -sqlServerAddress $installSettings.sqlServerAddress -installConfigPath $installConfigPath -quiet $quiet
 $identityDatabase = Get-IdentityDatabaseConnectionString -identityDbName $installSettings.identityDbName -sqlServerAddress $sqlServerAddress -installConfigPath $installConfigPath -quiet $quiet
 $metadataDatabase = Get-MetadataDatabaseConnectionString -metadataDbName $installSettings.metadataDbName -sqlServerAddress $sqlServerAddress -installConfigPath $installConfigPath -quiet $quiet
+
+# using methods in DosInstallUtilites to install idpss, which will make it easier to migrate the identity code later 
+$configStore = @{Type = "File"; Format = "XML"; Path = $installConfigPath}
+$idpssConfigStore = Get-DosConfigValues -ConfigStore $configStore -Scope "idpss"
+$commonConfigStore = Get-DosConfigValues -ConfigStore $configStore -Scope "common"
```
michaelvidal24: Should we move the above block closer to the idpss install section?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,64 @@ if ($fabricInstallerSecret){
...
+
+$idpssAppPoolUser = $idpssIISUser.UserName
+
+# uses Discovery url, we should be able to remove since I Get-IISAppPoolUser above
+$idpssName = "IdentityProviderSearchService"
+#$idpssAppPoolUser = Find-IISAppPoolUser -applicationName $idpssName -discoveryServiceUrl $discoveryServiceUrl -noDiscoveryService $true -quiet $quiet
+
+if($useAzure) {
+    Add-PermissionToPrivateKey $idpssAppPoolUser $selectedCerts.SigningCertificate read
+}
```
michaelvidal24: Should we run this block of code no matter what? I don't think it can hurt, and will make the logic more readable. We grant the user permission to read the private key so we can decrypt any secrets.

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,64 @@ if ($fabricInstallerSecret){
...
+
+$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($selectedSite.physicalPath), $idpssName)
+New-AppRoot $idpssDirectory $idpssAppPoolUser
+
+Add-DatabaseSecurity $idpssAppPoolUser $installSettings.identityDatabaseRole $identityDatabase.DbConnectionString
```
michaelvidal24: Does the idpss user need database access? I think this should only be done for the identity app pool user.

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,64 @@ if ($fabricInstallerSecret){
...
+Register-ServiceWithDiscovery -iisUserName $idpssAppPoolUser -metadataConnStr $metadataDatabase.DbConnectionString -version $idpssInstallApplication.version -serverUrl $idpssServiceUrl `
+-serviceName $idpssName -friendlyName "Fabric.IdentityProviderSearchService" -description "The Fabric.IdentityProviderSearchService searches Identity Providers for matching users and groups.";
+
+#$registrationApiSecret = Add-RegistrationApiRegistration -identityServerUrl $identityServiceUrl -accessToken $accessToken
+
+# uses discovery url similar to above
+#$idpssConfig = Get-WebConfigPath -applicationName $idpssName -discoveryServiceUrl $discoveryServiceUrl -noDiscoveryService $true -quiet $quiet
+$siteName = $idpssConfigStore.siteName
+$sitePath = $(get-item "iis:\sites\$siteName").physicalPath
+$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($sitePath), $idpssName)
```
michaelvidal24: Seems like we do something similar in Line 170 - can we just piggy back off that variable instead of recreating the path again?

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,64 @@ if ($fabricInstallerSecret){
...
+$idpssConfig = $idpssDirectory + "\web.config"
+
+Set-IdentityProviderSearchServiceWebConfigSettings -webConfigPath $idpssConfig `
```
michaelvidal24: Do we need this? For web deploy based powershell installs, we use the -webDeployParameters parameter of Publish-DosWebApplication which performs the web.config transform for us.

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +136,64 @@ if ($fabricInstallerSecret){
...
+$idpssInstallerPath = "..\WebDeployPackages\Fabric.IdentityProviderSearchService.zip"
+$idpssInstallPackagePath = Get-WebDeployPackagePath -standalonePath $idpssStandalonePath -installerPath $idpssInstallerPath
+
+$idpssIISUser = Get-IISAppPoolUser -credential $credential -appName $idpssConfigStore.appName -storedIisUser $idpssConfigStore.iisUser -installConfigPath $installConfigPath -scope $idpssSettingsScope
+
+$idpssInstallApplication = Publish-DosWebApplication -WebAppPackagePath $idpssInstallPackagePath `
+                      -WebDeployParameters $idpssWebDeployParameters `
+                      -AppPoolName $idpssConfigStore.appPoolName `
+                      -AppPoolCredential $idpssIISUser.credential `
+                      -AuthenticationType "Windows" `
```
michaelvidal24: IdPSS authentication should be `Anonymous`

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1168,6 +1178,114 @@ function Set-IdentityUri {
...
+        [Parameter(Mandatory=$true)]
+        [Hashtable] $serviceConfig,
+        [Parameter(Mandatory=$true)]
+        [Hashtable] $commonConfig
+    )
+
+    Confirm-ServiceConfig -commonConfig $commonConfig -serviceConfig $serviceConfig 
+    $metaDataConnectionString = Get-MetadataConnectionString -commonConfig $commonConfig -serviceConfig $serviceConfig
+    
+    $webDeployParameters = @(
```
michaelvidal24: I think we need the DiscoveryServiceEndPoint, UseDiscoveryService, and DomainName parameters in here.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: It is used in the WebDeployParameters

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: They are used in WebDeployParameters

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: sure.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: yes

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: sounds good.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: ok. thanks

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: yup. Looks like I had some duplication as I was moving code around.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: So do you mean we don't need to insert the azure app settings into the web.config. I believe that is part of our Identity and IdPSS azure registration script.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -462,6 +472,33 @@ function Add-RegistrationApiRegistration([string] $identityServerUrl, [string] $
     return $registrationApiSecret
 }
 
+function Add-IdpssApiResourceRegistration($identityServiceUrl, $fabricInstallerSecret)
+{
+    $accessToken = Get-AccessToken -authUrl $identityServiceUrl -clientId "fabric-installer" -secret $fabricInstallerSecret -scope fabric/identity.manageresources
+
+    Write-DosMessage -Level Information "Registering IdentitySearchProvider API with Fabric.Identity..."
+    $apiName = "identitySearchProvider-service-api"
```
michaelvidal24: I think today we register this with this name `idpsearch-api`

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -462,6 +472,33 @@ function Add-RegistrationApiRegistration([string] $identityServerUrl, [string] $
...
+    try{
+        [string]$apiSecret = [string]::Empty
+        Write-Host "    Registering $($apiName) with Fabric.Identity"
+        $body = New-APIRegistrationBody -apiName $apiName -userClaims @("name", "email", "roles", "group") -scopes @{"name" = "fabric/idprovider.searchusers"} -isEnabled $true
+        $apiSecret = New-ApiRegistration -identityUrl $identityServiceUrl -body (ConvertTo-Json $body) -accessToken $accessToken
+
+        if (![string]::IsNullOrEmpty($apiSecret) -and ![string]::IsNullOrWhiteSpace($apiSecret)) {
```
michaelvidal24: I think IsNullOrWhiteSpace is sufficient here, It also does a null check.

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -150,6 +129,11 @@ if ($fabricInstallerSecret){
     Write-DosMessage -Level "Information" -Message "Fabric.Installer clientSecret has been created."
 }
 
+# Call the Idpss powershell script
+.\Install-IdentityProviderSearchService.ps1 -certificates $($selectedCerts)
```
michaelvidal24: If we already know the credential, should we pass it to the IdPSS script? I think its OK for IdPSS and Identity to run as the same user.

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
```diff
@@ -0,0 +1,61 @@
...
+    [switch] $quiet
+)
+
+Import-Module -Name .\Install-Identity-Utilities.psm1 -Force
+
+# Get Idpss app pool user 
+# Create log directory with read/write permissions for app pool user
+# using methods in DosInstallUtilites to install idpss, which will make it easier to migrate the identity code later 
+$idpssSettingsScope = "identityProviderSearchService"
+$configStore = @{Type = "File"; Format = "XML"; Path = $installConfigPath}
```
michaelvidal24: Should we only do this if $configStore is null? Looks like we default it...so this would overwrite it.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: ok

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: yup.

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
hckenmiller: I think I can just remove this since that was before I set a default on the parameter above, since it wouldn't know what $installConfigPath was.

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
```diff
@@ -0,0 +1,85 @@
+﻿param(
+    [PSCredential] $credential, 
+    [Hashtable] $configStore = @{Type = "File"; Format = "XML"; Path = "$PSScriptRoot\install.config"},
+    [Hashtable] $certificates,
```
michaelvidal24: Instead of passing this in, can we get this from the common config?

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
```diff
@@ -0,0 +1,85 @@
...
+                      -WebDeploy `
+
+$idpssAppPoolUser = $credential.UserName
+
+$idpssName = "IdentityProviderSearchService"
+
+Add-PermissionToPrivateKey $idpssAppPoolUser $certificates.SigningCertificate read
+
+$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($selectedSite.physicalPath), $idpssName)
+New-AppRoot $idpssDirectory $idpssAppPoolUser
```
michaelvidal24: Can we rename this function to New-LogsDirectoryForApp or something similar? We aren't really creating a new application root here, since that was already done in the Publish step above.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1168,6 +1205,188 @@ function Set-IdentityUri {
...
+        [string] $settingName,
+        [string] $settingValue
+    )
+
+    if([string]::IsNullOrEmpty($settingValue)){
+        Write-DosMessage -Level "Fatal" -Message "You must specify a valid $settingName in config."
+    }
+}
+
+function Get-MetadataConnectionString{
```
michaelvidal24: I think we already have a function that does this: https://github.com/HealthCatalyst/Fabric.Identity/pull/299/files#diff-8cbfa091db649c4bbaf53cade86637cdR307

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
hckenmiller: I think it is the $certificates that you are asking about, if not let me know. The last method Set-IdentityProviderSearchServiceWebConfigSettings requires the SigningCertificate and the EncryptionCertificate.Thumbprint. We only have the primary and encryption thumbprints in common.config. The signing certificate looks like a version of the PrimarySigningCertificateThumbprint with a regex replace. I guess I could call Get-Certificates again, but I thought I tried that and had an issue.

Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
hckenmiller: Sure.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: I didn't add a new function for Get-MetadataConnectionString, there is still only one, I think it must just be a parenthesis change or something.

michaelvidal24: What is the difference between the function on line 307 vs the function on line 1310?

The one on line 307 is called `Get-MetadataDatabaseConnectionString` the one on line 1310 is called `Get-MetadataConnectionString`

PR #300, opened 2019-03-14
Add check to see if any switches were applied

joshnielsen10:
Check to see if user supplied either registerIdentity and/or registerIdPSS switch with this script. If not, display a message with examples to use.

### DIFF
--- a/Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
+++ b/Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
@@ -15,12 +15,11 @@ param(
 
 if(!($registerIdentity) -and !($registerIdPSS))
 {
-	Write-Host " You must register either Identity and/or IdPSS by adding their respective switch.
+	throw " You must register either Identity and/or IdPSS by adding their respective switch.
 	Examples:
 		.\Register-Identity-IdPSS.ps1 -registerIdentity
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Register-Identity-IdPSS.ps1
```diff
@@ -13,6 +13,16 @@ param(
     [switch] $registerIdPSS
 )
 
+if(!($registerIdentity) -and !($registerIdPSS))
+{
+	Write-Host " You must register either Identity and/or IdPSS by adding their respective switch.
+	Examples:
+		.\Register-Identity-IdPSS.ps1 -registerIdentity
+		.\Register-Identity-IdPSS.ps1 -registerIdPSS
+		.\Register-Identity-IdPSS.ps1 -registerIdentity -registerIdPSS" -ForegroundColor Red -BackgroundColor Black
```
ctediz: Can just do a Write-DosMessage -Level "Fatal" to avoid having to set colors, it should also throw an error which should cause the script to exit

joshnielsen10: Okay sounds good. I wasn't sure if I should throw an error or just exit. I changed it to `throw` which is basically shorthand for what you just described. Thanks for your input

PR #302, opened 2019-03-19
Remove progress bar when downloading assets during installation

rlipenta-hc:
Using the newly introduced "Get-WebRequestDownload" function in DosInstallUtilities to remove the progress bar when downloading assets for installation.

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -1,13 +1,5 @@
-# Import Fabric Install Utilities
-$fabricInstallUtilities = ".\Fabric-Install-Utilities.psm1"
-if (!(Test-Path $fabricInstallUtilities -PathType Leaf)) {
-    Write-DosMessage -Level "Warning" -Message "Could not find fabric install utilities. Manually downloading and installing"
-    Invoke-WebRequest -Uri https://raw.githubusercontent.com/HealthCatalyst/InstallScripts/master/common/Fabric-Install-Utilities.psm1 -Headers @{"Cache-Control" = "no-cache"} -OutFile $fabricInstallUtilities
-}
-Import-Module -Name $fabricInstallUtilities -Force
-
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/scripts/Install-Identity-Dev.ps1
```diff
@@ -38,7 +38,7 @@ function Create-CouchDb($couchDbUsername, $couchDbPassword, $couchDbServer)
 		
 		try{
 			Write-Host "Docker not installed, downloading and installing CouchDB for Windows..."
-			Invoke-WebRequest -Uri https://dl.bintray.com/apache/couchdb/win/2.1.0/apache-couchdb-2.1.0.msi -OutFile $env:Temp\apache-couchdb-2.1.0.msi
+			Get-WebRequestDownload -Uri https://dl.bintray.com/apache/couchdb/win/2.1.0/apache-couchdb-2.1.0.msi -OutFile $env:Temp\apache-couchdb-2.1.0.msi
```
hc-jordenlowe: Seeing the no caching option, seems like the structure of this function should follow similar to HttpClient, where you would pass a request object in, as opposed to all of these parameters.  Was that something you considered?  Im sure it would require more time, but could be expanded as new fields are needed.

Example idea:
```
$request = @{
  verb: "Get"
  url: "xxx",
  headers: [
    "cache-control": "no-cache"
  ],
  ...
}
Get-WebRequestDownload -request $request -OutFile $someFilePath
```

rlipenta-hc: There's certainly room to expand the behavior of Get-WebRequestDownload as it is initially a wrapper for Invoke-WebRequest. The NoCache flag is value-add, reducing the code the consumer has to write to utilize the wrapper. Long term I can see other value added flags get added, as I think there is more value it in providing value-add benefit than being a parameter for parameter wrapper of the existing Invoke-WebRequest.

rlipenta-hc:
Reverted changes to abandoned scripts as well as the Register scripts.

PR #304, opened 2019-03-22
Updates for the release pipeline

hckenmiller:


Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -136,7 +136,7 @@ if ($fabricInstallerSecret){
 }
 
 # Call the Idpss powershell script
-.\Install-IdentityProviderSearchService.ps1 -credential $iisUser.Credential -noDiscoveryService $noDiscoveryService -quiet $quiet
+.\Install-IdentityProviderSearchService.ps1 -credential $iisUser.Credential
```
michaelvidal24: I think we want to create an explicit config store here. since we are getting a config path in the parent script, we want to make sure that is our path to the config file otherwise we will get strange results.

PR #309, opened 2019-03-26
User/bsmithvt/login failure bug

bsmithvt:


Fabric.Identity.API/Quickstart/Account/AccountController.cs
```diff
@@ -258,6 +258,7 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl)
             }
             catch(InvalidIssuerException exc)
             {
+                await HttpContext.Authentication.SignOutAsync(IdentityServerConstants.ExternalCookieAuthenticationScheme);
```
michaelvidal24: Is this the only cookie that is set at this point that we have to delete?

bsmithvt: I believe so - I deleted the same cookie that is deleted at the end of this method after user login succeeds.

PR #310, opened 2019-03-27
Update scripts so that we don't get in a bad state in interactive ins…

michaelvidal24:
…tall if common config properties are not defined.

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -1228,10 +1228,12 @@ function Get-IdpssWebDeployParameters{
         [Parameter(Mandatory=$true)]
         [Hashtable] $commonConfig,
         [PSCredential] $credential,
+        [string] $applicationEndpoint,
         [string] $discoveryServiceUrl,
         [string] $noDiscoveryService,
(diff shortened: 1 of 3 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1282,8 +1284,6 @@ function Confirm-ServiceConfig{
 
     Confirm-SettingIsNotNull -settingName "common.sqlServerAddress" -settingValue $commonConfig.sqlServerAddress
     Confirm-SettingIsNotNull -settingName "common.metadataDbName" -settingValue $commonConfig.metadataDbName
-    Confirm-SettingIsNotNull -settingName "common.webServerDomain" -settingValue $commonConfig.webServerDomain
-    Confirm-SettingIsNotNull -settingName "common.clientEnvironment" -settingValue $commonConfig.clientEnvironment
```
ctediz: Was clientEnvironment being used for anything? Not seeing it being assigned in anything that was removed, but it looks like it was still being checked for like on this line

michaelvidal24: I could not find any references to it, so I removed it.

PR #312, opened 2019-03-28
Call the idpss script in quiet mode so we don't prompt for the same v…

michaelvidal24:
…alues 2x.

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity.ps1
@@ -36,7 +36,7 @@ if(!(Test-IsRunAsAdministrator))
 }
 
 $ErrorActionPreference = "Stop"
-
+$configStore = @{Type = "File"; Format = "XML"; Path = "$installConfigPath"}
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -130,8 +134,7 @@ if ($fabricInstallerSecret){
 }
 
 # Call the Idpss powershell script
-$configStore = @{Type = "File"; Format = "XML"; Path = "$installConfigPath"}
-.\Install-IdentityProviderSearchService.ps1 -credential $iisUser.Credential -configStore $configStore -noDiscoveryService:$noDiscoveryService -quiet:$quiet
+.\Install-IdentityProviderSearchService.ps1 -credential $iisUser.Credential -configStore $configStore -noDiscoveryService:$noDiscoveryService -quiet
```
ctediz: Looks like the script is prompting for everything in the identity install script if not quiet, and always calling the idpss install as quiet.

Looks reasonable

PR #316, opened 2019-04-01
added logic to handle previously unhandled exception

bsmithvt:


Fabric.Identity.API/Quickstart/Account/AccountController.cs
```diff
@@ -262,12 +262,21 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl)
...
+            try
+            {
+                //issue authentication cookie for user
+                user = await _userLoginManager.UserLogin(
+                    claimInformation.Provider,
+                    _claimsService.GetEffectiveUserId(claimInformation),
+                    claimInformation.Claims,
+                    claimInformation?.ClientId);
+            }
+            catch (Exception e)
```
michaelvidal24: Should we be catching a generic exception here? Seems like we might cover up some issues that we would want to know about and fix.

bsmithvt: Yes, this is for that bug. Are there any cases where we want the error to be surfaced to the end user? I've also added more details to the log - I just noticed the actual details of the SqlException are due to data being potentially truncated.

michaelvidal24: I don't think we can just swallow it, we need to figure out how to handle this exceptional case so users can successfully authenticate.

bsmithvt: Are you saying they should be able to authenticate in this scenario? Why would I be able to authenticate with my HC account if it's not whitelisted?

michaelvidal24: I believe the HC tenant is a whitelisted tenant in the exploratory test environment.

PR #319, opened 2019-04-02
bypass any group claims that start with [ to fix SQL error during aut…

bsmithvt:
…hentication

### DIFF
--- a/Fabric.Identity.API/Management/UserLoginManager.cs
+++ b/Fabric.Identity.API/Management/UserLoginManager.cs
@@ -6,8 +6,6 @@ using System.Threading.Tasks;
 using Fabric.Identity.API.Models;
 using Fabric.Identity.API.Persistence;
 using IdentityModel;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using Serilog;
 
 namespace Fabric.Identity.API.Management
--- a/Fabric.Identity.API/Management/UserLoginManager.cs
+++ b/Fabric.Identity.API/Management/UserLoginManager.cs
@@ -6,6 +6,8 @@ using System.Threading.Tasks;
 using Fabric.Identity.API.Models;
 using Fabric.Identity.API.Persistence;
 using IdentityModel;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Serilog;
 
 namespace Fabric.Identity.API.Management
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/Management/UserLoginManager.cs
```diff
@@ -113,7 +113,17 @@ private List<Claim> FilterClaims(IEnumerable<Claim> incomingClaims)
                 // copy the claim as-is
                 else
                 {
-                    filtered.Add(claim);
+                    if (claim.Type == FabricIdentityConstants.FabricClaimTypes.Groups)
+                    {
+                        if (string.IsNullOrEmpty(claim.Value) || !claim.Value.StartsWith("["))
```
michaelvidal24: Do we want to add empty claims to the list?

Fabric.Identity.API/Management/UserLoginManager.cs
```diff
@@ -113,7 +113,17 @@ private List<Claim> FilterClaims(IEnumerable<Claim> incomingClaims)
                 // copy the claim as-is
                 else
                 {
-                    filtered.Add(claim);
+                    if (claim.Type == FabricIdentityConstants.FabricClaimTypes.Groups)
```
michaelvidal24: Should we add some of these nested if's to an else if to reduce some nesting?

Fabric.Identity.API/Management/UserLoginManager.cs
bsmithvt: We were adding them in the original version.

```
else
{
   filtered.Add(claim);
}
```

Fabric.Identity.API/Management/UserLoginManager.cs
bsmithvt: Do you mean pulling the `if (claim.Type == FabricIdentityConstants.FabricClaimTypes.Groups)` out of the `else`?

Fabric.Identity.API/Management/UserLoginManager.cs
michaelvidal24: ok - just looked weird...what claims do we get that are empty?

bsmithvt: I don't see any in my DB that are null or empty, but the `UserClaims` table `Value` column is nullable.

Fabric.Identity.API/Management/UserLoginManager.cs
michaelvidal24: Yeah, basically combining it with the else above to make it an else if - but its not really necessary.

Fabric.Identity.API/Management/UserLoginManager.cs
michaelvidal24: ok

PR #323, opened 2019-04-10
created global variable representing IdPSS app name and use that in s…

bsmithvt:
…cripts


ctediz:
Looks like it's possible to export a variable instead of setting a global like this:
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/export-modulemember?view=powershell-6#description

bsmithvt:
@ctediz, I'll read up on it later, but do you know off hand if one approach preferred over or are there any differences?

ctediz:
> @ctediz, I'll read up on it later, but do you know off hand if one approach preferred over or are there any differences?

Thinking of it now I'm not sure, since module seem to stay in scope so it may end up being the same in the end.
@miketabet Thoughts?

PR #329, opened 2019-04-29
Removed Links in Identity Login

joshnielsen10:
Removed the links at the footer of the Identity login since they pointed nowhere.

### DIFF
--- a/Fabric.Identity.API/Views/Account/Login.cshtml
+++ b/Fabric.Identity.API/Views/Account/Login.cshtml
@@ -47,10 +47,6 @@
                                     </a>
                                 </div>
                             }
-                            <div id="login-footer">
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/Views/Account/Login.cshtml
```diff
@@ -84,24 +80,5 @@
                 </div>
             </div>
             <br><br><br>
-            <div class="footerWrapper" align="center">
-                <div class="footerBottom" align="center">
```
hc-jordenlowe: Not that we need it, but what if we leave these divs and only remove the links below?

joshnielsen10: Ya I'll put them back for a possible future use.

PR #331, opened 2019-05-02
created error page to use for Fabric.Identity errors

bsmithvt:


Fabric.Identity.API/Quickstart/Home/HomeController.cs
```diff
@@ -62,5 +62,10 @@ public async Task<IActionResult> Error(string errorId)
 
             return View("Error", vm);
         }
+
+        public IActionResult UnauthorizedError()
```
hc-jordenlowe: Do you need to add this to the web.config too?  like a default 401 setting there?  I might be thinking of the old stuff.

bsmithvt: @hc-jordenlowe, yeah, I initially tried that approach but switched to this.

PR #341, opened 2019-05-21
refactored additional claims logic so that Azure AD tokens contain gi…

bsmithvt:
…ven_name, family_name, and email claims

### DIFF
--- a/Fabric.Identity.API/Models/ClaimsResult.cs
+++ b/Fabric.Identity.API/Models/ClaimsResult.cs
@@ -18,5 +18,22 @@ namespace Fabric.Identity.API.Models
         public string SchemeItem { get; set; }
         public List<Claim> Claims { get; set; }
         public Claim UserIdClaim { get; set; }
+
+        public override string ToString()
+        {
+            var sb = new System.Text.StringBuilder();
+
+            foreach (var c in Claims)
+            {
+                sb.Append($"Type={c.Type}, Value={c.Value}, ");
+            }
+
+            foreach (var c in AdditionalClaims)
+            {
+                sb.Append($"Type={c.Type}, Value={c.Value}, ");
+            }
+
+            return $"Provider={Provider}, UserId={UserId}, ClientId={ClientId}, SchemeItem={SchemeItem}, {sb} ";
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


Fabric.Identity.API/Models/ClaimsResult.cs
```diff
@@ -18,5 +18,22 @@ public class ClaimsResult
...
+            var sb = new System.Text.StringBuilder();
+
+            foreach (var c in Claims)
+            {
+                sb.Append($"Type={c.Type}, Value={c.Value}, ");
+            }
+
+            foreach (var c in AdditionalClaims)
+            {
+                sb.Append($"Type={c.Type}, Value={c.Value}, ");
```
hc-jordenlowe: is the comma's at the end going to cause issues on the last element?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -95,37 +94,57 @@ public string GetEffectiveUserId(ClaimsResult claimInformation)
...
+            var claims = claimsPrincipal.Claims.ToList();
             var userIdClaim = this.GetUserIdClaim(claims);
-            return new ClaimsResult()
+
+            var subjectId = AzureADSubjectId(claims);
+
+            if (!string.IsNullOrEmpty(subjectId))
+            {
+                var externalUser = await _externalIdentityProviderService.FindUserBySubjectId(subjectId);
+                if (externalUser != null)
```
hc-jordenlowe: This is the second time I seen this code.  Could we create a common function of it?

Fabric.Identity.API/Services/ClaimsService.cs
```diff
@@ -136,8 +155,9 @@ private ClaimsResult GenerateNewClaimsResult(AuthenticateInfo info, Authorizatio
             };
         }
 
-        private Claim[] GenerateAdditionalClaims(List<Claim> previousClaims)
+        private Claim[] GenerateAdditionalClaims(ClaimsResult claimsResult, AuthenticateInfo authenticateInfo)
```
hc-jordenlowe: AuthenticateInfo authenticateInfo parameter is not used.  Is this required?  Maybe we remove the need for this?

PR #342, opened 2019-05-24
Use a filter to find graph api permissions

ctediz:
Do not require PSCredentials (let the azureAD module handle that)

### DIFF
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -85,13 +85,14 @@ function New-FabricAzureADApplication() {
         [Microsoft.Open.AzureAD.Model.RequiredResourceAccess] $permission,
         [bool] $isMultiTenant = $false
     )
+    $groupMembershipClaims = 'SecurityGroup'
 
     $app = Get-AzureADApplication -Filter "DisplayName eq '$appName'" -Top 1
     if($null -eq $app) {
-        $app = New-AzureADApplication -Oauth2AllowImplicitFlow $true -RequiredResourceAccess $permission -DisplayName $appName -ReplyUrls $replyUrls.name -AvailableToOtherTenants $isMultiTenant
+        $app = New-AzureADApplication -Oauth2AllowImplicitFlow $true -RequiredResourceAccess $permission -DisplayName $appName -ReplyUrls $replyUrls.name -AvailableToOtherTenants $isMultiTenant -GroupMembershipClaims $groupMembershipClaims
     }
     else {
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -160,12 +160,15 @@ function Connect-AzureADTenant {
     param(
         [Parameter(Mandatory=$true)]
         [string] $tenantId,
-        [Parameter(Mandatory=$true)]
         [PSCredential] $credential
     )
 
     try {
-        Connect-AzureAD -Credential $credential -TenantId $tenantId | Out-Null
+        if($credential) {
```
ctediz: Remove the mandatory flag.  On the HC tenant, we use SSO, that redirect was causing the login to fail.  
If not credential is provided, then the AzureAD module handles that authentication portion, by opening up in a browser window which handles that case better.

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -21,7 +21,7 @@ else {
 
 function Get-GraphApiDirectoryReadPermissions() {
     try {
-        $aad = @((Get-AzureADServicePrincipal | Where-Object {$_.ServicePrincipalNames.Contains("https://graph.microsoft.com")}))[0]
+        $aad = @((Get-AzureADServicePrincipal -Filter "ServicePrincipalNames eq 'https://graph.microsoft.com'"))[0]
```
ctediz: The HC tenant had a lot of applications registered. The results where getting paged and the graph api was not being found on that first page.  This filter should only select the graph api application instead of grabbing all applications and then searching for it in the response.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -26,7 +26,7 @@ try{
 }
 Import-Module -Name CatalystDosIdentity -MinimumVersion $minVersion -Force
 
-$Global:idPSSAppName = "IdentityProviderSearchService"
+$Global:idPSSAppName = "Identity Provider Search Service"
```
ctediz: Switching to spaces to match documenatation/identity service registration

bsmithvt: I feel like I changed this at some point because it was broken? I remember introducing the global variable, so maybe the bug was the fact that it was mismatched in 2 different scripts, so as long as it's the same everywhere it's fine.

ctediz: Ya there was a bug in the past as there was one script using spaces for the name, and another not using spaces. 
Changing it to spaces since it matches what the docs currently have for the examples.

PR #343, opened 2019-05-29
Use the install.config appname for registration portions

ctediz:


Fabric.Identity.API/scripts/Install-IdentityProviderSearchService.ps1
```diff
@@ -60,19 +60,17 @@ $idpssWebDeployParameters = Get-IdpssWebDeployParameters -serviceConfig $idpssCo
...
                       -WebDeploy
 
-$idpssName = $Global:idPSSAppName
-
-$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($selectedSite.physicalPath), $idpssName)
+$idpssDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($selectedSite.physicalPath), $idpssConfigStore.appName)
 Write-Host "IdPSS Directory: $($idpssDirectory)"
 New-LogsDirectoryForApp $idpssDirectory $idpssIisUser.UserName
 
 Register-ServiceWithDiscovery -iisUserName $idpssIisUser.UserName -metadataConnStr $metadataDatabase.DbConnectionString -version $idpssInstallApplication.version -serverUrl $idpssServiceUrl `
```
miketabet: While it's not part of your fix, we may want to consider adding some logic to validate the IdPSSUrl is "good" as part of this code review.  I think we've seen several incidents due to URL issues with bad URLs in discovery service.  Also, we may want to do some basic validation on the IdPSS service account.

ctediz: We already do validation on the service account further up in the script in the 'Get-IISAppPoolUser' portion.

Did you have some thoughts on how to verify the idpssurl is valid? Where you thinking to try and ping the service with that url before trying to post or similar?

PR #344, opened 2019-06-02
Added Default 401

joshnielsen10:
Added a default 401 error page for Identity.

### DIFF
--- a/Fabric.Identity.API/Startup.cs
+++ b/Fabric.Identity.API/Startup.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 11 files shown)


Fabric.Identity.API/Startup.cs
```diff
@@ -201,6 +202,18 @@ public void ConfigureServices(IServiceCollection services)
             }
 
             app.UseExceptionHandler("/Home/UnauthorizedError");
+            app.UseStatusCodePages(
```
ctediz: Is it better to use UseStatusCodePagesWithRedirects? It seems like the status code might change to a 302 with that though.  What happens to the status code in this case?

joshnielsen10: I looked at that as well. Using UseStatusCodePagesWithRedirects never shows you had a 401 in the dev tools. It just redirects like nothing was wrong. The better one was UseStatusCodePagesWithReExecute but that wouldn't work. It caused more issues and sometimes when clicking the Windows button to login, instead of seeing the login, it would just throw a 500 error. Using UseStatusCodePages is similar to the ReExecute just more code. You still see the 401 status code in the dev tools with UseStatusCodePages.

PR #345, opened 2019-06-03
User/bsmithvt/group claim header fix

bsmithvt:


Fabric.Identity.API/Services/GroupFilterService.cs
```diff
@@ -23,23 +23,33 @@ public IEnumerable<Claim> FilterClaims(IEnumerable<Claim> claims)
...
+            }
+
             var claimList = claims.ToList();
-            var prefixFilteredClaims = claimList;
-            var suffixFilteredClaims = claimList;
+            var prefixFilteredClaims = new List<Claim>();
+            var suffixFilteredClaims = new List<Claim>();
 
-            if (prefixes?.Count > 0)
+            if (prefixFilters?.Count > 0)
```
miketabet: Consider putting in some logging so we can quickly identify if these filters are working.  Perhaps log the claim count before and after applying the filters?

PR #350, opened 2019-06-07
User/jason/handleduplicateregistrations

ctediz:
Update scripts to handle multiple registrations

Append to registered application ReplyUrls instead of replacing.
Require a secret name through install config to identify the secret in the registration
Prompt for secrets to be deleted if duplicate names are found

Still needs to be tested

### DIFF
--- a/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
@@ -93,10 +93,12 @@ function New-FabricAzureADApplication() {
     }
     else {
         # Do not overwrite, append to existing urls
+        # Updating app fails if trying to add duplicate urls
         $existingUrls = $app.ReplyUrls
         foreach($replyUrl in $replyUrls.name) {
             $existingUrls.Add($replyUrl)
         }
+        $existingUrls = $existingUrls | Select-Object -Unique
         Set-AzureADApplication -ObjectId $app.ObjectId -RequiredResourceAccess $permission -Oauth2AllowImplicitFlow $true -ReplyUrls $existingUrls -AvailableToOtherTenants $isMultiTenant -GroupMembershipClaims $groupMembershipClaims
     }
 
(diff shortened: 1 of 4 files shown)


Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -314,6 +349,16 @@ function Confirm-Tenants {
     }
 }
 
+function Get-InstallIdPSSUtilsUserConfirmation {
+    Write-DosMessage -Level "Information" -Message "Found ${$filterKeys.count} duplicate secrets."
+    $deleteSecrets = Read-Host  "Delete duplicate secret(s)? [Y/N]"
+    switch ($deleteSecrets) {
+        Y {return $true}
```
hc-jordenlowe: is this taking into account lower case Y as well?

ctediz: Yes, this defaults to a case insentive check

Fabric.Identity.API/scripts/Install-IdPSS-Utilities.psm1
```diff
@@ -92,13 +92,18 @@ function New-FabricAzureADApplication() {
         $app = New-AzureADApplication -Oauth2AllowImplicitFlow $true -RequiredResourceAccess $permission -DisplayName $appName -ReplyUrls $replyUrls.name -AvailableToOtherTenants $isMultiTenant -GroupMembershipClaims $groupMembershipClaims
     }
     else {
-        Set-AzureADApplication -ObjectId $app.ObjectId -RequiredResourceAccess $permission -Oauth2AllowImplicitFlow $true -ReplyUrls $replyUrls.name -AvailableToOtherTenants $isMultiTenant -GroupMembershipClaims $groupMembershipClaims
+        # Do not overwrite, append to existing urls
+        $existingUrls = $app.ReplyUrls
+        foreach($replyUrl in $replyUrls.name) {
+            $existingUrls.Add($replyUrl)
```
ctediz: If a duplicate replyUrl is added, when the call is made to update the application, then the update fails.  Need to find a way to filter out existing replyUrls

PR #351, opened 2019-06-12
removed validation from install-discovery script

hc-jordenlowe:
Weird bug.  If you call this individually, then it validates the hashtable.

If you call it from the main script, it works with no validation.


hc-jordenlowe:
> Looks good - the issue is that you're validating a string with your decorators but $_ refers to a hashtable which doesn't work with Test_Path ($_ refers to $configStore, not $configStore.Path). If you want to validate path, you need to use $_.Path

neat, i can add that in then.

hc-jordenlowe:
looks like there is an issue with hash tables.  Instead of the validation being done in the parameter in an aspect oriented fashion, i will just add the code to the first lines of the function.

PR #360, opened 2019-07-17
Add v1 to idpss registration to discovery

ctediz:


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -386,8 +386,12 @@ function Get-ApplicationEndpoint([string] $appName, [string] $applicationEndpoin
             $defaultAppEndpoint = $userEnteredApplicationEndpoint
         }
     }
-    if($defaultAppEndpoint){ Add-InstallationSetting $scope "applicationEndPoint" "$defaultAppEndpoint" $installConfigPath | Out-Null }
-    if($defaultAppEndpoint){ Add-InstallationSetting "common" "$($scope)Service" "$defaultAppEndpoint" $installConfigPath | Out-Null }
+
+    # Do not need to add IdPSS to common config, only used by Identity
+    if($defaultAppEndpoint -notlike "*IdentityProviderSearchService*"){
```
bsmithvt: Is this string in a constant somewhere?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -386,8 +386,12 @@ function Get-ApplicationEndpoint([string] $appName, [string] $applicationEndpoin
...
+        if($defaultAppEndpoint){ Add-InstallationSetting $scope "applicationEndPoint" "$defaultAppEndpoint" $installConfigPath | Out-Null }
+        if($defaultAppEndpoint){ Add-InstallationSetting "common" "$($scope)Service" "$defaultAppEndpoint" $installConfigPath | Out-Null }
```
bsmithvt: Is `common` in a constant somewhere?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
miketabet: We should move this up as a global variable, or even better make it a parameter with a default value of "IdenitytProviderSearchService".

miketabet: Make IdentityProviderSearchService a variable - even better if you can make it a parameter with the default value set to "IdentityProviderSearchService", but allow overrides.

ctediz: maybe a parameter/flag it check against before writing to the install.config instead?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: there doesn't seem to be a constant for that in this script

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: does idpss still need to have the variable applicationEndpoint created for the Idpss scope. if($defaultAppEndpoint){ Add-InstallationSetting $scope "applicationEndPoint" "$defaultAppEndpoint" $installConfigPath | Out-Null }

ctediz: Let me verify, but I'm pretty sure nothing uses it, it's just in there because we're reusing the identity Get-ApplicationEndpoint which adds it in by default

ctediz: Ya nothing references this anywhere from what I can see

hckenmiller: ok.

PR #362, opened 2019-07-26
User/jason/seperateinstallconfig

ctediz:
Reference separate azure setting config for Identity/IdPSS AAD install

Added some extra validation for AAD settings

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Discovery-IdPSS.ps1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Discovery-IdPSS.ps1
@@ -15,7 +15,7 @@ param(
         return $true
     })] 
     [string] $installConfigPath = "$PSScriptRoot\install.config", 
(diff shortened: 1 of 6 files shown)


Fabric.Identity.API/scripts/Install-Identity-Discovery-IdPSS.ps1
```diff
@@ -15,6 +15,7 @@ param(
         return $true
     })] 
     [string] $installConfigPath = "$PSScriptRoot\install.config", 
+    [string] $azureConfigPath = "C:\Program Files\Health Catalyst\azuresettings.config",    # TODO: env variable for program files?
```
ctediz: Forgot to look into an environment variable or similar for the program files path, will try to find something

joshnielsen10: According to Mr. Tabet, it's $env:ProgramFiles

joshnielsen10: According to Mr. Tabet, it's $env:ProgramFiles

PR #363, opened 2019-07-29
add webserverdomain code

hckenmiller:
this code adds a default webServerDomain that prevents powershell script failure

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -296,6 +296,22 @@ function Get-SqlServerAddress([string] $sqlServerAddress, [string] $installConfi
     return $sqlServerAddress
 }
 
+function Get-WebServerDomain([string] $webServerDomain, [string] $installConfigPath, [bool] $quiet){
+    if([string]::IsNullOrEmpty($webServerDomain))
+	{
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -296,6 +296,22 @@ function Get-SqlServerAddress([string] $sqlServerAddress, [string] $installConfi
     return $sqlServerAddress
 }
 
+function Get-WebServerDomain([string] $webServerDomain, [string] $installConfigPath, [bool] $quiet){
+    if([string]::IsNullOrEmpty($webServerDomain))
```
miketabet: There's the case for vanity URLs that we may want to consider detecting/warning about in the default scenario, but not necessarily worth holding up this fix.

hckenmiller: So if they enter a url instead of a webserverdomain, we could check for https:// or www.?

miketabet: Sorry, I think we'll need to remove this check, as I think the webserver domain should not include the protocol header

hckenmiller: it doesn't, this is a check to make sure it doesn't. I thought that was what you were referring to. I will remove it.

PR #367, opened 2019-08-09
added code for migrating AAD settings from install.config to azuresettings.config

hckenmiller:
I am currently working to do testing in DosInstaller to see that it is functioning as expected.


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+    }
+
+    # Gather AAD Settings from install.config
+    $existingChildNodes = Get-XMLChildNodes -installConfigPath $installConfigPath -configSection $configSection -nodesToSearch $nodesToSearch -childNodeGetAttribute $childNodeGetAttribute
+
+    # Remove blank child nodes from default azuresettings.config
+    Remove-XMLChildNodes -azureConfigPath $azureConfigPath -configSection $configSection -nodesToSearch $nodesToSearch -childNodeGetAttribute $childNodeGetAttribute -childNodeAttributeSetting ""
+
+    # Add AAD settings from Doc1
+    Add-XMLChildNodesFromDoc1 -azureConfigPath $azureConfigPath -configSection $configSection -doc1ChildNodesInOrder $nodesToSearch -doc1ChildNodesToAdd $existingChildNodes
```
ctediz: Can this just be renamed to 'Add-XMLChildNodes', and change the parameters a bit as well?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+        [Parameter(Mandatory=$true)]
+        [string] $azureConfigPath,
+        [Parameter(Mandatory=$true)]
+        [string] $configSection,
+        [Parameter(Mandatory=$true)]
+        [string[]] $nodesToSearch,
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeGetAttribute,
+        [string] $childNodeAttributeSetting,
+        [string] $reverseOperator
```
ctediz: seems like this can be a boolean of some kind, but I'm having a hard time coming up with a name

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+    )
+    # Clean out empty variables in default azureSettings.config
+    $azureInstallationConfig = [xml](Get-Content $azureConfigPath)
+    $azureIdentityScope = $azureInstallationConfig.installation.settings.scope | Where-Object {$_.name -eq $configSection}
+
+    foreach($nodeToSearch in $nodesToSearch)
+    {
+     $setAzureSettings = $azureIdentityScope.SelectSingleNode($nodeToSearch)
+
+     if($reverseOperator -eq "not equal")
```
ctediz: if I'm understanding this, if this is set to "not equal" then it removes all other nodes otherwise it removes only the nodesToSearch's children?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+      $allExistingChildNodes += $existingChildNodes
+      if($null -eq $existingChildNodes)
+      {
+        Write-DosMessage -Level "Information" -Message "$($nodeToSearch) node or child attributes not found"
+      }
+    }
+    return $allExistingChildNodes
+}
+
+function Check-XMLChildNodes {
```
ctediz: Should this also check for azureSecretName?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+    )
+
+    $installationConfig = [xml](Get-Content $installConfigPath)
+    $identityScope = $installationConfig.installation.settings.scope | Where-Object {$_.name -eq $configSection}
+    $existingChildNode = @{}
+
+    $existingChildNode = $identityScope.ChildNodes | Where-Object {$_.$childNodeGetAttribute -eq $childNodeAttributeSetting}
+    if ($null -eq $existingChildNode)
+    {
+      Write-DosMessage -Level "Information" -Message "$childNodeAttributeSetting not found in install.config"
```
ctediz: Should not finding the node be concerning? Or is it ok to just continue, like not finding azureSecretName for example

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+    $existingChildNode = @{}
+
+    $existingChildNode = $azureIdentityScope.ChildNodes | Where-Object {$_.$childNodeGetAttribute -eq $childNodeAttributeSetting}
+    if ($null -ne $existingChildNode)
+    {
+      $existingChildNode.$childNodeSetAttribute = $childNodeAttributeValue
+      $azureInstallationConfig.Save("$azureConfigPath") | Out-Null
+      Write-DosMessage -Level "Information" -Message "$childNodeAttributeValue added to $childNodeAttributeSetting in azureSettings.config"
+    }
+    elseif ($null -eq $existingChildNode)
```
ctediz: At this point this is just setting values right? So if the node doesn't have a value, then it just stays as empty right?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: sure

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: yah. I need to clean that up.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: If it is set to "" then it removes the default azuresettings.config <variable name = "" alias = ""/> for each nodeToSearch. If it is set to "not equal" then it removes the <variable name="something" alias = "something" /> currently in install.config, after migrating settings to azuresettings.config for each nodeToSearch.

hckenmiller: If the reverseoperator is not set, then it removes the default azuresettings.config blank <variable name = "" alias = ""/> for each nodeToSearch, where $childNodeAttributeSetting -eq "". If the reverseoperator is set to "not equal" then it removes the non-blank <variable name="something" alias = "something" /> currently in install.config, for each nodeToSearch, where $childNodeAttributeSetting -ne "".

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: In the current use case, I would hope to find azureSecretName, so that would be an issue. The function is written though to be used more generically, so it could be used for other code.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: Since the parameter is mandatory, $childNodeAttributeValue would have to be passed, not empty string, for the function to not fail.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,303 @@ function Test-DiscoveryService{
...
+        [string] $installConfigPath,
+        [Parameter(Mandatory=$true)]
+        [string] $azureConfigPath
+    )
+    $configSection = "identity"
+    $nodesToSearch = @("tenants", "replyUrls", "claimsIssuerTenant", "allowedTenants", "registeredApplications")
+    $childNodeGetAttribute = "name"
+    $childNodeSetAttribute = "value"
+
+    Write-DosMessage -Level "Information" -Message "Get the AAD Setting contents in install.config"
```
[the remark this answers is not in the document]
hckenmiller: It is more to know where the migration process got to, if we had to troubleshoot. Maybe I will change the wording to be AAD Setting check complete, so the user doesn't think there is an issue.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: Since the azureSecretName variable is contained at the identity scope level, I am just using the Get-XMLChildNode and Set-XMLChildNode for that, so I don't have to do more if checks to use or not use the SelectSingleNode xml function.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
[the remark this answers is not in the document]
miketabet: The other option is to move this to verbose.  In general, take an logs that are too chatty for the typical end user and move them to debug or verbose since they still have lots of value for troubleshooting.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+    param (
+        [Parameter(Mandatory=$true)]
+        [string] $installConfigPath,
+        [Parameter(Mandatory=$true)]
+        [string] $configSection,
+        [Parameter(Mandatory=$true)]
+        [string] $nodeToSearch,
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeGetAttribute,
+        [string] $childNodeGetAttribute2
```
miketabet: Comment why we need attribute and attribute2.  Use case isn't super clear for me.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+    $childNodeGetAttribute = "name"
+    $childNodeSetAttribute = "value"
+
+    Write-DosMessage -Level "Information" -Message "Get the AAD Setting contents in install.config"
+    
+    # Quick check to see if any AAD settings are present in install.config (short circuit)
+    $existingAADSettings = Check-XMLChildNode -installConfigPath $installConfigPath -configSection $configSection -nodeToSearch $nodesToSearch[0] -childNodeGetAttribute $childNodeGetAttribute
+    if($false -eq $existingAADSettings)
+    {
+      Write-DosMessage -Level "Information" -Message "AAD Setting check complete for install.config"
```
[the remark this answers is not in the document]
hckenmiller: That is what it was before the change based on @ctediz comment

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+
+    # Get and Set AAD azureSecretName 
+    $azureSecretName = Get-XMLChildNode -installConfigPath $installConfigPath -configSection $configSection -childNodeGetAttribute $childNodeGetAttribute -childNodeAttributeSetting "azureSecretName" 
+    Set-XMLChildNode -azureConfigPath $azureConfigPath -configSection $configSection -childNodeGetAttribute $childNodeGetAttribute -childNodeSetAttribute $childNodeSetAttribute -childNodeAttributeSetting "azureSecretName" -childNodeAttributeValue $azureSecretName.value
+ 
+    # Gather AAD Settings from azuresettings.config
+    $existingAzureChildNodes = Get-XMLChildNodes -installConfigPath $azureConfigPath -configSection $configSection -nodesToSearch $nodesToSearch -childNodeGetAttribute $childNodeGetAttribute
+    
+    # Check if install.config and azuresettings.config AAD settings match
+    $areSame = Compare-Object -ReferenceObject $existingChildNodes -DifferenceObject $existingAzureChildNodes -Property name, alias, appName, tenantId, tenantAlias, clientid, secret
```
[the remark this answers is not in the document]
hckenmiller: It is deep if you are using the correct -Property.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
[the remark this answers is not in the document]
hckenmiller: Did that, not seeing them logged after setting minimumLoggingLevel = "Verbose"

Fabric.Identity.API/scripts/Install-Identity-Discovery-IdPSS.ps1
```diff
@@ -39,6 +39,26 @@ $discoveryScope = "discoveryservice"
 $discoveryConfig = Get-DosConfigValues -ConfigStore $configStore -Scope $discoveryScope
 $enableOAuth = [string]::IsNullOrEmpty($discoveryConfig.enableOAuth) -ne $true -and $discoveryConfig.enableOAuth -eq "true"
 
+# Check if install.config has AAD settings and azureConfigStore is empty
+# update azureConfigStore with the newly created config file.
+$existingAzurePath = Test-Path $azureConfigPath -PathType Leaf
+if($false -eq $existingAzurePath)
+{
+  $updatedInstallConfigPath = "$env:ProgramFiles\Health Catalyst\install.config"
```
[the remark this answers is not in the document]
hckenmiller: ok.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeGetAttribute,
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeAttributeSetting
+    )
+    # Clean out empty variables in default azureSettings.config
+    $azureInstallationConfig = [xml](Get-Content $azureConfigPath)
+    $azureIdentityScope = $azureInstallationConfig.installation.settings.scope | Where-Object {$_.name -eq $configSection}
+
+    $removeEmptyVariable = $azureIdentityScope.ChildNodes | Where-Object {$_.$childNodeGetAttribute -eq $childNodeAttributeSetting}
```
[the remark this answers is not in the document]
hckenmiller: I will add that.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+        [Parameter(Mandatory=$true)]
+        [string] $azureConfigPath,
+        [Parameter(Mandatory=$true)]
+        [string] $configSection,
+        [Parameter(Mandatory=$true)]
+        [string[]] $nodesToSearch,
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeGetAttribute,
+        [string] $childNodeAttributeSetting,
+        [switch] $negateOperator
```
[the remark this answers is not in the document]
hckenmiller: It changes the variable check from looking at variables that are empty in azuresettings.config to those that aren't in install.config. Example of empty: "< variable name = "" alias = "" />" Example of not empty: "< variable name = "something" alias = "whatever" />"

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,305 @@ function Test-DiscoveryService{
...
+        [Parameter(Mandatory=$true)]
+        [string[]] $nodesToSearch,
+        [Parameter(Mandatory=$true)]
+        [string] $childNodeGetAttribute
+    )
+
+    $installationConfig = [xml](Get-Content $installConfigPath)
+    $identityScope = $installationConfig.installation.settings.scope | Where-Object {$_.name -eq $configSection}
+    $allExistingChildNodes = @()
+    $count = 1
```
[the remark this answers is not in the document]
hckenmiller: I have to add a place holder for each node, so I know when one ends and another one starts. I might be able to leave off the count, since I'm not using that in the check.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: Since I wanted to make the function more generic, when checking for existing child nodes. A check on the name attribute works for the tenants, but not the registered applications, since appName could be the same but tenantId and appName is unique.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,335 @@ function Test-DiscoveryService{
...
+    $existingChildNodes = Get-XMLChildNodes -installConfigPath $installConfigPath -configSection $configSection -nodesToSearch $nodesToSearch -childNodeGetAttribute $childNodeGetAttribute
+    $azureSecretNameFromInstallConfig = Get-XMLChildNode -installConfigPath $installConfigPath -configSection $configSection -childNodeGetAttribute $childNodeGetAttribute -childNodeAttributeSetting "azureSecretName" 
+    $childNodeHeader = @{}
+    $childNodeHeader.Add("secretName","")
+    $existingChildNodes += $childNodeHeader
+    $existingChildNodes += $azureSecretNameFromInstallConfig
+    Write-DosMessage -Level "Verbose" -Message "Copied the AAD Settings from install.config"
+
+    # Remove blank child nodes and azureSecretName from default azuresettings.config manifest
+    Write-DosMessage -Level "Verbose" -Message "Started removing blank child variables and azureSecretName from azuresettings.config in nuget folder"
```
ctediz: This might not always be in the nuget folder if a migrationAzureConfigPath parameter is passed to the script right?
Would writing the azureConfigPath instead of 'nuget folder' be better here?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1625,6 +1625,335 @@ function Test-DiscoveryService{
...
+    $existingAzureChildNodes = Get-XMLChildNodes -installConfigPath $azureConfigPath -configSection $configSection -nodesToSearch $nodesToSearch -childNodeGetAttribute $childNodeGetAttribute
+    Write-DosMessage -Level "Verbose" -Message "Copied the AAD Settings from azuresettings.config for comparison"
+
+    # Check if install.config and azuresettings.config AAD settings match
+    $areSameGroup = Compare-Object -ReferenceObject $existingChildNodes -DifferenceObject $existingAzureChildNodes -Property name, value, alias, appName, tenantId, tenantAlias, clientid, secret
+    Write-DosMessage -Level "Verbose" -Message "Comparing the AAD Settings in install.config and azuresettings.config, before deleting from install.config"
+
+    if($null -ne $areSameGroup)
+    {
+      Write-DosMessage -Level "Verbose" -Message "Since install.config and azuresettings.config AAD Settings are different, install.config settings were not deleted"
```
ctediz: Probably want this line in "Informational" or higher so it comes up even when not debugging, since it sounds like something went wrong somewhere

Fabric.Identity.API/scripts/azuresettings.config
```diff
@@ -19,7 +19,9 @@
         <variable name="" alias="" />
       </allowedTenants>
       <!-- The name of the secret to be created for the azure application registration -->
+      <secretName>
```
ctediz: Do the registration scripts need to change because of this as well?

Fabric.Identity.API/scripts/Install-Identity-Discovery-IdPSS.ps1
```diff
@@ -39,6 +41,36 @@ $discoveryScope = "discoveryservice"
 $discoveryConfig = Get-DosConfigValues -ConfigStore $configStore -Scope $discoveryScope
 $enableOAuth = [string]::IsNullOrEmpty($discoveryConfig.enableOAuth) -ne $true -and $discoveryConfig.enableOAuth -eq "true"
 
+# Check if install.config has AAD settings and azureConfigStore is empty
+# update azureConfigStore with the newly created config file.
+$existingAzurePath = Test-Path $azureConfigPath -PathType Leaf
+if($false -eq $existingAzurePath)
+{
+  $existingInstallPath = Test-Path $migrationInstallConfigPath -PathType Leaf
```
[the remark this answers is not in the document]
hckenmiller: In the installer, this script is currently invoked without modifying parameters, just uses the defaults, so I can't modify the config files that I need unless I include the other two locations.

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: Yah, only if it is run outside of the DosInstaller.

Fabric.Identity.API/scripts/azuresettings.config
hckenmiller: added change to registration script

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: changed

PR #378, opened 2019-09-19
User/jason/182979 installintegration

ctediz:
Integrate the identity secret handling changes into the identity install script

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -2006,7 +2006,6 @@ function Test-IdentityEncryptionCertificateValid {
     param(
         [System.Security.Cryptography.X509Certificates.X509Certificate2] $encryptionCertificate
     )
-    # Update a month before
     $today = Get-Date
     return $encryptionCertificate.NotAfter -gt $today
 }
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -2099,10 +2099,12 @@ function Get-IdentityFabricInstallerSecret {
     if ($fabricInstallerSecret.StartsWith("!!enc!!:")) {
         $secretNoEnc = $fabricInstallerSecret -replace "!!enc!!:"
         $fabricInstallerSecret = Unprotect-DosInstallerSecret -CertificateThumprint $encryptionCertificateThumbprint -EncryptedInstallerSecretValue $secretNoEnc
-        if ([string]::IsNullOrWhitespace($fabricInstallerSecret)) {
-            # create new secret if no secret or unable to decrypt
-            $fabricInstallerSecret = Invoke-ResetFabricInstallerSecret -identityDbConnectionString $identityDbConnectionString
-        }
+    }
+
+    # Create new secret if one does not exist, or was unable to decrypt
+    if ([string]::IsNullOrWhitespace($fabricInstallerSecret)) {
+        # create new secret if no secret or unable to decrypt
+        $fabricInstallerSecret = Invoke-ResetFabricInstallerSecret -identityDbConnectionString $identityDbConnectionString
     }
 
     return $fabricInstallerSecret
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -269,7 +269,7 @@ function Add-PermissionToPrivateKey([string] $iisUser, [System.Security.Cryptogr
             Write-DosMessage -Level "Fatal" -Message "No key file was found at '$($cspKeyPath)' or '$($cngKeyPath)' for '$($signingCert)'. Ensure a valid signing certificate was provided"
         }
 
-        $acl = Get-Acl $cspKeyPath
+        $acl = Get-Acl $keyPath
         $acl.AddAccessRule($allowRule)
         Set-Acl $keyPath $acl -ErrorAction Stop
         Write-DosMessage -Level "Information" -Message "The permission '$($permission)' was successfully added to the private key for user '$($iisUser)'"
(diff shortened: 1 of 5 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -1982,30 +1982,77 @@ function Remove-FilePermissions
...
     return $encryptionCertificate.NotAfter -gt $today
 }
 
 function Remove-IdentityEncryptionCertificate {
     param(
-        [string] $encryptionCertificateThumbprint
+        [string] $encryptionCertificateThumbprint,
+        [string] $friendlyName = "Fabric Identity Signing Encryption Certificate"
     )
-    $cert = Get-Certificate $encryptionCertificateThumbprint
```
ctediz: Fun note: Authorization uses this function as well, this causes the Authorization install.config to have a signing/encryption certificate in it's own install.config section.

Authorization will need to be updated to either not use this, or to grab the encryption certificate value from the common section of the install.config

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -71,14 +68,40 @@ $currentDirectory = $PSScriptRoot
...
-$iisUser = Get-IISAppPoolUser -credential $credential -appName $installSettings.appName -storedIisUser $installSettings.iisUser -installConfigPath $configStore.Path -scope $installSettingsScope
-Add-PermissionToPrivateKey $iisUser.UserName $selectedCerts.SigningCertificate read
 $appInsightsKey = Get-AppInsightsKey -appInsightsInstrumentationKey $installSettings.appInsightsInstrumentationKey -installConfigPath $configStore.Path -scope $installSettingsScope -quiet $quiet
 $sqlServerAddress = Get-SqlServerAddress -sqlServerAddress $commonInstallSettings.sqlServerAddress -installConfigPath $configStore.Path -quiet $quiet
 $identityDatabase = Get-IdentityDatabaseConnectionString -identityDbName $installSettings.identityDbName -sqlServerAddress $sqlServerAddress -installConfigPath $configStore.Path -quiet $quiet
 $metadataDatabase = Get-MetadataDatabaseConnectionString -metadataDbName $commonInstallSettings.metadataDbName -sqlServerAddress $sqlServerAddress -installConfigPath $configStore.Path -quiet $quiet
 
+# Switch to DosInstallUtilities later, testing with Fabric-Install-Utilities locally
```
ctediz: Seems like a good section to move into its own function and add tests around

Fabric.Identity.API/scripts/Install-Identity.ps1
```diff
@@ -71,14 +68,40 @@ $currentDirectory = $PSScriptRoot
...
+    Remove-IdentityEncryptionCertificate -encryptionCertificateThumbprint $commonInstallSettings.encryptionCertificateThumbprint
+    $encryptionCertificate = New-IdentityEncryptionCertificate
+}
+
+# Wait until potentially creating new certificate to add permissions to private key
+$iisUser = Get-IISAppPoolUser -credential $credential -appName $installSettings.appName -storedIisUser $installSettings.iisUser -installConfigPath $configStore.Path -scope $installSettingsScope
+Add-PermissionToPrivateKey $iisUser.UserName $encryptionCertificate read
+
+# update install.config
+# Assumes both encryption and signing cert are the same certificate
```
ctediz: Important note here ⏫

ctediz: Currently, the only way to separate the encryption/signing certificate is to do so manually through the install.config.  
The Get-Certificates behavior assumed that these would be the same as well.
With this new implementation, everytime the new cert is generated, then both the encryption/signing certs are replaced.

Is this assumption still valid?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
hckenmiller: I like having the common stuff like encryption certificate in common section.

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: I always remembered them showing the same value, and it seemed weird to me that they are the same but separated out.

Fabric.Identity.API/scripts/tests/Remove-IdentityEncryptionCertificate.tests.ps1
```diff
@@ -0,0 +1,100 @@
...
+                $thumbPrint = "non-existing-thumbprint"
+                Mock -ModuleName Install-Identity-Utilities -CommandName Write-DosMessage -MockWith {}
+                Mock -ModuleName Install-Identity-Utilities -CommandName Remove-Item -MockWith {}
+
+                # Act
+                Remove-IdentityEncryptionCertificate -encryptionCertificateThumbprint $thumbPrint
+
+                # Assert
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName Write-DosMessage -Times 1 -Exactly -Scope It
+                Assert-MockCalled -ModuleName Install-Identity-Utilities -CommandName Write-DosMessage -ParameterFilter { $Level -and $Level -eq "Information" -and $Message.StartsWith("Certificate with thumbprint") } -Times 1 -Exactly -Scope It
```
hckenmiller: Does the message also contain "not found"?

Fabric.Identity.API/scripts/tests/Remove-IdentityEncryptionCertificate.tests.ps1
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        InModuleScope Install-Identity-Utilities {
+            It 'Should not attempt to remove when certificate is not the identity encryption certificate' {
+                # Arrange
+                $testCert = New-Object -TypeName System.Security.Cryptography.X509Certificates.x509Certificate2
+                $friendlyNameProperty = @{
+                    MemberType = [System.Management.Automation.PSMemberTypes]::NoteProperty
+                    Name = "FriendlyName"
+                    value = "Some Friendly Name"
```
hckenmiller: So the value in the friendlyNameProperty would have to be "Fabric Identity Signing Encryption Certificate" for it to be removed?

Fabric.Identity.API/scripts/Install-Identity.ps1
hckenmiller: yah.

Fabric.Identity.API/scripts/tests/Remove-IdentityEncryptionCertificate.tests.ps1
ctediz: Ya, it defaults to "Fabric Identity Signing Encryption Certificate" to look for, but that can be passed in to be something else.

The idea was to not accidently remove the SSL certificate, or any certificate that we did not create ourselves.

Fabric.Identity.API/scripts/tests/Remove-IdentityEncryptionCertificate.tests.ps1
ctediz: Ya, i can make the message to take for be more explicit

PR #390, opened 2019-10-18
User/bskeen/update to dotnet core 2

bskeen:
Upgrading Fabric.Identity to .Net Core 2.2 to allow for IdPSS functionality to be merged into Fabric.Identity.


Fabric.Identity.API/Validation/ApiResourceValidator.cs
```diff
@@ -22,9 +22,14 @@ private void ConfigureRules()
 
             RuleFor(apiResource => apiResource.Scopes)
                 .NotNull()
-                .SetCollectionValidator(new ScopeValidator())
                 .WithMessage("Please specify at least one Scope for this Api Resource");
 
+            RuleForEach(apiResource => apiResource.Scopes)
+                .SetValidator(new ScopeValidator())
+                .WithMessage("Please specify at least one Scope for this Api Resource");
```
ctediz: Is this rule being duplicated?

Fabric.Identity.API/appsettings.json
```diff
@@ -26,9 +26,9 @@
     "Password": ""
   },
   "HostingOptions": {
-    "UseIis": false,
-    "UseTestUsers": true,
-    "StorageProvider": "InMemory",
+    "UseIis": true,
+    "UseTestUsers": false,
+    "StorageProvider": "sqlserver",
```
ctediz: I'm thinking this file should be undone, but it seems like we are always making these changes to get it to run locally

Fabric.Identity.IntegrationTests/SuppressExecutionContextHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Fabric.Identity.IntegrationTests
+{
+    public class SuppressExecutionContextHandler : DelegatingHandler
+    {
+        public SuppressExecutionContextHandler(HttpMessageHandler innerHandler) : base(innerHandler)
+        {
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
ctediz: What is this class doing?

Fabric.Identity.API/Events/FabricUserLoginSuccessEvent.cs
```diff
@@ -4,12 +4,9 @@ namespace Fabric.Identity.API.Events
 {
     public class FabricUserLoginSuccessEvent : UserLoginSuccessEvent
     {
-        public FabricUserLoginSuccessEvent(string provider, string providerUserId, string subjectId, string name, string clientId) : base(provider, providerUserId, subjectId, name)
+        public FabricUserLoginSuccessEvent(string provider, string providerUserId, string subjectId, string name, string clientId)
```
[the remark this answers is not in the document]
bskeen: This one was already that way, it was just a formatting change here. This class is sent to IdentityServer, and the type name may be used to log some stuff.

Fabric.Identity.API/Extensions/AuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+                            options.Scope.Add(s);
+                        }
+                    });
+                }
+            }
+
+            return builder;
+        }
+
+        public static IApplicationBuilder UseStaticFilesForAcmeChallenge(this IApplicationBuilder builder, string challengeDirectory, ILogger logger)
```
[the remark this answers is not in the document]
bskeen: Yes, this particular one is. It is still used in the Configure method.

Fabric.Identity.API/Validation/ApiResourceValidator.cs
bskeen: No, it's not. One of the validators checks if the collection is null, and this one checks if there are items in the collection and that those items are valid.

Fabric.Identity.IntegrationTests/SuppressExecutionContextHandler.cs
bskeen: Without this class (and the use of it in the Integration Test Fixture), the integration tests all break in the middleware. The TestServer clears the HttpClient from the HttpContext, so when the middleware tries to access it, the server crashes with a Null Reference Exception. This is a known breaking change caused by .Net Core 2.2.

Fabric.Identity.API/scripts/BenTestLog.txt
```diff
@@ -0,0 +1,4772 @@
+Error response from daemon: No such container: identity.integration.openldap
```
ctediz: extra file

Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -1,161 +1,161 @@
...
+            _appConfig = appConfig;
+            _logger = logger;
+            _policyProvider = policyProvider;
+            _httpClient = httpClient;
+            _httpRequestMessageFactory = httpRequestMessageFactory;
+        }
+
+        public async Task<ExternalUser> FindUserBySubjectId(string subjectId)
+        {
+            if (!_appConfig.IdentityProviderSearchSettings.IsEnabled)
```
ctediz: Is this part of the port over?

Seems strange that only this search has a check to see if it's disabled, it's probably better to inject an empty IExternalIdentityProviderService during startup

Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
```diff
@@ -1,161 +1,161 @@
...
+            };
+
+            var accessTokenResponse = await client.RequestClientCredentialsTokenAsync(tokenRequest);
+            if (accessTokenResponse.IsError)
+            {
+                _logger.Error(
+                    $"Failed to get access token, error message is: {accessTokenResponse.ErrorDescription}");
+            }
+
+            var baseUri = await _appConfig.IdentityProviderSearchSettings.GetEffectiveBaseUrl(_appConfig);
```
ctediz: Is this trying to reach out to old idpss? I'm guessing this is existing code?

ctediz: I'm guessing this is handled in the other pr actually

Fabric.Identity.API/Documentation/SecurityRequirementsOperationFilter.cs
```diff
@@ -19,13 +19,24 @@ public SecurityRequirementsOperationFilter(IOptions<AuthorizationOptions> author
...
+            if (context.ApiDescription.ActionDescriptor is
+                Microsoft.AspNetCore.Mvc.Controllers.ControllerActionDescriptor controller)
+            {
+                policies.AddRange(controller.ControllerTypeInfo
+                    .GetCustomAttributes(true)
+                    .OfType<AuthorizeAttribute>()
+                    .Select(attr => attr.Policy));
+            }
+
+            if (context.ApiDescription.TryGetMethodInfo(out var methodInfo))
```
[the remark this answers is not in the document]
hckenmiller: It looks like you could get the policy names that determines access to the resource, by consolidating to this and add a .Select(attr => attr.Policy) at the end.         
var authAttributes = context.MethodInfo.DeclaringType.GetCustomAttributes(true)
            .Union(context.MethodInfo.GetCustomAttributes(true))
            .OfType<AuthorizeAttribute>();
https://github.com/domaindrivendev/Swashbuckle.AspNetCore#operation-filters

Fabric.Identity.API/Fabric.Identity.API.csproj
```diff
@@ -1,23 +1,29 @@
...
-    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
     <DockerComposeProjectPath>..\docker-compose.dcproj</DockerComposeProjectPath>
     <Company>Health Catalyst</Company>
     <Product>Fabric.Identity</Product>
     <Copyright>Copyright © 2018</Copyright>
     <Version Condition=" '$(BUILD_BUILDNUMBER)' == '' ">1.0.0.0-local</Version>
     <Version Condition=" '$(BUILD_BUILDNUMBER)' != '' ">$(BUILD_BUILDNUMBER)</Version>
+    <CopyRefAssembliesToPublishDirectory>true</CopyRefAssembliesToPublishDirectory>
+    <AspNetCoreHostingModel>OutOfProcess</AspNetCoreHostingModel>
```
hckenmiller: Why are we choosing OutOfProcess instead of InProcess for hosting in Asp.Net Core 2.2. Is it because we want to be compatible with Linux as well so InProcess won't work for linux. InProcess provides better performance and is less resource intensive. There isn't an extra network hop between IIS and Kestral or maintaining an additional process.

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -58,8 +58,10 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     PostLogoutRedirectUris = client.PostLogoutRedirectUris,
                     EnableLocalLogin = client.EnableLocalLogin,
                     IdentityProviderRestrictions = client.IdentityProviderRestrictions,
-                    LogoutUri = client.LogoutUri,
-                    LogoutSessionRequired = client.LogoutSessionRequired,
+                    FrontChannelLogoutUri = client.LogoutUri,
+                    BackChannelLogoutUri = client.LogoutUri,
```
hckenmiller: We specify GrantTypes.Implicit or implicit flow above, but we are using BackChannelLogoutUri and LogoutSessionRequired, which is the hybrid flow here.

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -76,7 +78,7 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     Claims = client.Claims,
                     AlwaysSendClientClaims = client.AlwaysSendClientClaims,
                     AlwaysIncludeUserClaimsInIdToken = client.AlwaysIncludeUserClaimsInIdToken,
-                    PrefixClientClaims = client.PrefixClientClaims,
+                    ClientClaimsPrefix = client.PrefixClientClaims ? "client_" : null,
```
hckenmiller: Was the prefixclientclaims set to "client_" before? Is this just to distinguish client claims from user claims?

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -120,8 +122,8 @@ public static Client ToClientViewModel(this IS4.Client client)
                 PostLogoutRedirectUris = client.PostLogoutRedirectUris,
                 EnableLocalLogin = client.EnableLocalLogin,
                 IdentityProviderRestrictions = client.IdentityProviderRestrictions,
-                LogoutUri = client.LogoutUri,
-                LogoutSessionRequired = client.LogoutSessionRequired,
+                LogoutUri = client.FrontChannelLogoutUri ?? client.BackChannelLogoutUri,
```
hckenmiller: So we aren't using the hybrid, we are using either implicit flow or authorization code flow?

Fabric.Identity.API/Startup.cs
```diff
@@ -192,13 +213,12 @@ public void ConfigureServices(IServiceCollection services)
...
             IApplicationBuilder app,
             IHostingEnvironment env,
-            ILoggerFactory loggerFactory,
             IDbBootstrapper dbBootstrapper)
         {
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
-                _loggingLevelSwitch.MinimumLevel = LogEventLevel.Verbose;
+                LogFactory.LoggingLevelSwitch.MinimumLevel = LogEventLevel.Verbose;
```
hckenmiller: should this be something exposed in web.config or appsettings.json to allow the type of logging?

Fabric.Identity.IntegrationTests/IntegrationTestsFixture.cs
```diff
@@ -175,35 +185,83 @@ protected static IdentityDbContext IdentityDbContext
             }
         }
 
-        public HttpClient HttpClient { get; }
+        private readonly TaskCompletionSource<HttpClient> _httpClientTaskCompletionSource;
+
+        public Task<HttpClient> HttpClient => _httpClientTaskCompletionSource.Task;
 
         private TestServer CreateIdentityTestServer(string storageProvider)
         {
```
hckenmiller: Can this test server setup be consolidated some, and create the TestServer with a method and only the custom settings for each test server should be different.

Fabric.Identity.UnitTests/Validation/StubTokenValidator.cs
```diff
@@ -17,6 +19,11 @@ public Task<TokenValidationResult> ValidateAccessTokenAsync(string token, string
             bool validateLifetime = true)
         {
             return Task.FromResult(IdentityTokenValidationResult);
-        }
+        }
+
+        public Task<TokenValidationResult> ValidateRefreshTokenAsync(string token, Client client = null)
```
hckenmiller: why are parameters being passed, but not used?

Fabric.Identity.API/Fabric.Identity.API.csproj
bskeen: I chose to use OutOfProcess hosting because that is the way we are currently hosting it. My goal was to update the framework without having to drastically change how things are working. If this is the wrong call in this case, we can definitely have the discussion to update this, especially since I agree that it would be better if we were able to host InProcess.

Fabric.Identity.API/Models/ModelExtensions.cs
bskeen: With this one, the LogoutUri wasn't split into Front and Back channel before, so Jason and I decided to assign it to both. We weren't exactly sure which one we were using, and we were thinking it might be easier to put the same value in both to preserve the current functionality. We could remove one if it isn't needed though.

Fabric.Identity.API/Documentation/SecurityRequirementsOperationFilter.cs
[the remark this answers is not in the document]
bskeen: Thanks for the feedback, I will update this.

Fabric.Identity.API/Models/ModelExtensions.cs
bskeen: When we were updating IdentityServer, there were quite a few places where they updated types on objects we were using. In this case, it used to be a boolean value, but now it is a string. Looking up in the former code, it added `client_` as the prefix if the boolean was set to true. Although the new structure is more flexible, our code was built to work with the old structure. In particular, we had a Client model class that had a boolean for this parameter. The ternary operator above is there to try and accommodate the new structure in IdentityServer without rewriting the functionality we currently have.

Fabric.Identity.API/Models/ModelExtensions.cs
bskeen: Same as the comment above, we weren't sure which was used by our system. It could be that BackChannel is never used, we just were updating to try and preserve the current functionality as closely as possible (and before there was only one URL). I would be fine removing the null coalesce if one is never used by us.

Fabric.Identity.API/Services/IdPSearchServiceProvider.cs
bskeen: This class will probably be removed in the next couple of versions. The current search endpoint does reference this, although it might be broken (which is one of the reasons why we wanted to migrate IdPSS over in the first place).

Fabric.Identity.API/Startup.cs
bskeen: This should already be controlled by the configuration. In this case, it is just setting the logging to the highest detail level if it is in development mode.

Fabric.Identity.IntegrationTests/IntegrationTestsFixture.cs
bskeen: Thanks for the input, I fixed this and will push the changes soon.

Fabric.Identity.IntegrationTests/ServiceTests/IdentityConfigurationProviderTests.cs
```diff
@@ -23,10 +25,14 @@ public void GetAppConfiguration_ReturnsAppConfig_Success()
...
             var decryptionService = new DecryptionService(mockCertificateService);
-            var identityConfigurationProvider = new IdentityConfigurationProvider();
+            var config = new ConfigurationBuilder()
+                .AddJsonFile(Path.Combine(directory.FullName, "appsettings.json"))
+                .Build();
+            var identityConfigurationProvider = new IdentityConfigurationProvider(config);
+            
 
             // Act
-            var appConfig = identityConfigurationProvider.GetAppConfiguration(directory.FullName, decryptionService);
```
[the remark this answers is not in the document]
bskeen: This is related to the changes I needed to make to `Program.cs`'s main method. Configuration is handled a little bit differently in .Net Core 2.2 than in .Net Core 1.1.

Fabric.Identity.UnitTests/Validation/StubTokenValidator.cs
bskeen: I think that an interface was updated, and in this case, where this is a StubTokenValidator just used for testing purposes, I implemented the method, but without any real content in the method.

Fabric.Identity.IntegrationTests/IntegrationTestsFixture.cs
```diff
@@ -433,7 +463,9 @@ protected virtual void Dispose(bool disposing)
             if (disposing)
             {
                 // free managed resources
-                HttpClient.Dispose();
+                //HttpClient.Wait();
```
hckenmiller: So HttpClient can be removed now.

Fabric.Identity.IntegrationTests/IntegrationTestsFixture.cs
```diff
@@ -214,49 +259,34 @@ private TestServer CreateRegistrationApiTestServer(string storageProvider)
...
-            apiBuilder.UseKestrel()
-                .UseContentRoot(Directory.GetCurrentDirectory())
-                .UseStartup<Startup>()
-                .UseUrls(RegistrationApiServerUrl);
-
-            return new TestServer(apiBuilder);
+            builder.ConfigureServices(c => c.AddSingleton(options)
+                .AddSingleton<IPostConfigureOptions<JwtBearerOptions>>(new PostConfigureOptions<JwtBearerOptions>("Bearer", jwtOpts =>
+                {
+                    jwtOpts.BackchannelHttpHandler = new SuppressExecutionContextHandler(IdentityTestServer.CreateHandler());
```
hckenmiller: What is this doing?

PR #395, opened 2019-10-24
Migrate IdPSS functionality to Fabric.Identity

bskeen:


Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -58,8 +58,10 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     PostLogoutRedirectUris = client.PostLogoutRedirectUris,
                     EnableLocalLogin = client.EnableLocalLogin,
                     IdentityProviderRestrictions = client.IdentityProviderRestrictions,
-                    LogoutUri = client.LogoutUri,
-                    LogoutSessionRequired = client.LogoutSessionRequired,
+                    FrontChannelLogoutUri = client.LogoutUri,
```
hckenmiller: made comments on previous pull request which covers aspnet core 2.2 upgrade

Fabric.Identity.API/Fabric.Identity.API.csproj
```diff
@@ -1,23 +1,29 @@
...
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
     <DockerComposeProjectPath>..\docker-compose.dcproj</DockerComposeProjectPath>
     <Company>Health Catalyst</Company>
     <Product>Fabric.Identity</Product>
     <Copyright>Copyright © 2018</Copyright>
     <Version Condition=" '$(BUILD_BUILDNUMBER)' == '' ">1.0.0.0-local</Version>
     <Version Condition=" '$(BUILD_BUILDNUMBER)' != '' ">$(BUILD_BUILDNUMBER)</Version>
+    <CopyRefAssembliesToPublishDirectory>true</CopyRefAssembliesToPublishDirectory>
```
hckenmiller: made comment on previous pull request that covered aspnetcore 2.2 upgrade

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -76,7 +78,7 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     Claims = client.Claims,
                     AlwaysSendClientClaims = client.AlwaysSendClientClaims,
                     AlwaysIncludeUserClaimsInIdToken = client.AlwaysIncludeUserClaimsInIdToken,
-                    PrefixClientClaims = client.PrefixClientClaims,
```
hckenmiller: made comments on previous pull request which covers aspnet core 2.2 upgrade

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbResourceStore.cs
```diff
@@ -41,7 +41,7 @@ public Task<ApiResource> FindApiResourceAsync(string name)
             return DocumentDbService.GetDocument<ApiResource>(name);
```
hckenmiller: Can't we remove anything dealing with couchdb, since I don't believe we are using it as a store?

hckenmiller: It looks like a lot of this code is part of the previous pull request for aspnetcore 2.2 upgrade. I made comments in that pull request dealing with the upgrade.

hckenmiller:
A lot of this code is part of the pull request for aspnetcore 2.2 upgrade, my comments for that are in that pull request.

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbResourceStore.cs
hckenmiller: I didn't think we used CouchDbStore anymore, can we remove couchdb or is it still used somewhere?

Fabric.Identity.API/Services/Azure/AzureActiveDirectoryClientCredentialsService.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Fabric.Identity.API;
+using Fabric.Identity.API.Exceptions;
+using IdentityModel.Client;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Fabric.Identity.API.Configuration;
+using Microsoft.Extensions.Localization;
+
```
hckenmiller: It looks like most of this code for IdPSS functionality is directly pulled from there.

Fabric.Identity.API/Extensions/AuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+                        options.SignInScheme = IdentityServerConstants.ExternalCookieAuthenticationScheme;
+                        options.SignOutScheme = IdentityServerConstants.SignoutScheme;
+
+                        options.Authority = externalIdProvider.Authority;
+                        options.ClientId = externalIdProvider.ClientId;
+                        options.ClientSecret = externalIdProvider.ClientSecret;
+                        options.ResponseType = externalIdProvider.ResponseType;
+                        options.CallbackPath = "/signin-oidc-" + externalIdProvider.ProviderName;
+                        options.SignedOutCallbackPath = "/signout-callback-oidc-" + externalIdProvider.ProviderName;
+                        options.SaveTokens = true; // Okta needs the id_token for successful logout
```
johnrulon: Does this need to be a configuration value per external IdP? Okta needs it to be true, but could others need it false? I guess I'd be fine if we keep it true for now and let usage show us if we need to add it as a config.

Fabric.Identity.API/scripts/BenTestLog.txt
```diff
@@ -0,0 +1,4772 @@
+Error response from daemon: No such container: identity.integration.openldap
```
hckenmiller: Does this need to be checked in?

Fabric.Identity.API/web.config
```diff
@@ -12,13 +12,40 @@
...
+        <environmentVariable name="HostingOptions__StorageProvider" value="SqlServer" />
+        <environmentVariable name="SigningCertificateSettings__UseTemporarySigningCredential" value="false" />
+        <environmentVariable name="ConnectionStrings__IdentityDatabase" value="Server=HC2353.hqcatalyst.local;Database=Identity;Trusted_Connection=True;MultipleActiveResultSets=True;" />
+        <environmentVariable name="IdentityServerConfidentialClientSettings__Authority" value="http://localhost/IdentityDev" />
+        <environmentVariable name="SigningCertificateSettings__EncryptionCertificateThumbprint" value="EBC1D4A73E8061E78F2157C3D618544A464C299E" />
+        <environmentVariable name="DiscoveryServiceEndpoint" value="https://HC2353.hqcatalyst.local/DiscoveryService/v1" />
+        <environmentVariable name="AllowLocalLogin" value="false" />
+        <environmentVariable name="IdentityServerApiSettings__ApiSecret" value="secret" />
+        <environmentVariable name="IdentityServerConfidentialClientSettings__ClientSecret" value="secret" />
+        <environmentVariable name="AzureActiveDirectorySettings__Scope__1" value="profile" />
```
hckenmiller: All but the top three environmentvariables, shouldn't be checked in.

Fabric.Identity.API/Extensions/AuthenticationBuilderExtensions.cs
hckenmiller: The external IdPs don't need it, in my testing, it could have an if block, but we don't really know what they will call it for the providername, since the customer has to name it.

Fabric.Identity.API/Program.cs
```diff
@@ -1,25 +1,70 @@
...
+                    config.AddDockerSecrets(typeof(IAppConfiguration));
+                    config.SetBasePath(hostContext.HostingEnvironment.ContentRootPath);
+                })
+                .UseUrls("http://*:5001")
+                .ConfigureKestrel((context, options) =>
+                {
+                    // Set properties and call methods on options
+                })
+                .UseSerilog()
+                .UseStartup<Startup>();
```
johnrulon: Is all of this just auto added after the .NET Core upgrade? I'm not sure if all of it needed/ok to have. Maybe it's all fine.

Fabric.Identity.API/Quickstart/Account/AccountOptions.cs
```diff
@@ -21,7 +21,9 @@ public class AccountOptions
         public static bool WindowsAuthenticationEnabled = true;
         public static bool IncludeWindowsGroups = true;
         // specify the Windows authentication schemes you want to use for authentication
-        public static readonly string[] WindowsAuthenticationSchemes = new string[] { "Negotiate", "NTLM" };
+        // In aspnetcore 2.2 the GetAllSchemesAsync method doesn't return NTLM or Negotiate
+        // for windows authentication, the only one returned now is Windows.
+        public static readonly string WindowsAuthenticationSchemes = "Windows";
```
johnrulon: Does "Windows" mean it will accept both "Negotiate" and "NTLM"?

Fabric.Identity.API/Resource.resx
```diff
@@ -0,0 +1,126 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
```
johnrulon: Yay! A resource file! Nice.

Fabric.Identity.API/Quickstart/Account/AccountOptions.cs
hckenmiller: I believe so, that is the only option in AspNetCore 2.2

Fabric.Identity.API/Documentation/SecurityRequirementsOperationFilter.cs
```diff
@@ -19,13 +19,24 @@ public SecurityRequirementsOperationFilter(IOptions<AuthorizationOptions> author
...
-            var controllerPolicies = context.ApiDescription.ControllerAttributes()
-                .OfType<AuthorizeAttribute>()
-                .Select(attr => attr.Policy);
-            var actionPolicies = context.ApiDescription.ActionAttributes()
-                .OfType<AuthorizeAttribute>()
-                .Select(attr => attr.Policy);
-            var policies = controllerPolicies.Union(actionPolicies).Distinct();
+            List<string> policies = new List<string>();
+
+            if (context.ApiDescription.ActionDescriptor is
```
hckenmiller: It looks like you could get the policy names that determines access to the resource, by consolidating to this and add a .Select(attr => attr.Policy) at the end.         
       var authAttributes = context.MethodInfo.DeclaringType.GetCustomAttributes(true)
            .Union(context.MethodInfo.GetCustomAttributes(true))
            .OfType&lt;AuthorizeAttribute&gt;(); 
https://github.com/domaindrivendev/Swashbuckle.AspNetCore#operation-filters

Fabric.Identity.API/Fabric.Identity.API.csproj
```diff
@@ -1,23 +1,29 @@
...
+    <AspNetCoreHostingModel>OutOfProcess</AspNetCoreHostingModel>
```
hckenmiller: Why are we choosing OutOfProcess instead of InProcess for hosting in Asp.Net Core 2.2. Is it because we want to be compatible with Linux as well so InProcess won't work for linux? InProcess provides better performance and is less resource intensive. There isn't an extra network hop between IIS and Kestral or maintaining an additional process.

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -58,8 +58,10 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     PostLogoutRedirectUris = client.PostLogoutRedirectUris,
                     EnableLocalLogin = client.EnableLocalLogin,
                     IdentityProviderRestrictions = client.IdentityProviderRestrictions,
-                    LogoutUri = client.LogoutUri,
-                    LogoutSessionRequired = client.LogoutSessionRequired,
+                    FrontChannelLogoutUri = client.LogoutUri,
+                    BackChannelLogoutUri = client.LogoutUri,
```
hckenmiller: We specify that the AllowedGrantTypes for a Client is Implicit, which is FrontChannel. Why we using BackChannel?

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -76,7 +78,7 @@ public static IS4.Client ToIs4ClientModel(this Client client)
                     Claims = client.Claims,
                     AlwaysSendClientClaims = client.AlwaysSendClientClaims,
                     AlwaysIncludeUserClaimsInIdToken = client.AlwaysIncludeUserClaimsInIdToken,
-                    PrefixClientClaims = client.PrefixClientClaims,
+                    ClientClaimsPrefix = client.PrefixClientClaims ? "client_" : null,
```
hckenmiller: Was the prefixclientclaims set to "client_" before? Is this just to distinguish client claims from user claims?

Fabric.Identity.API/Models/ModelExtensions.cs
```diff
@@ -120,8 +122,8 @@ public static Client ToClientViewModel(this IS4.Client client)
                 PostLogoutRedirectUris = client.PostLogoutRedirectUris,
                 EnableLocalLogin = client.EnableLocalLogin,
                 IdentityProviderRestrictions = client.IdentityProviderRestrictions,
-                LogoutUri = client.LogoutUri,
-                LogoutSessionRequired = client.LogoutSessionRequired,
+                LogoutUri = client.FrontChannelLogoutUri ?? client.BackChannelLogoutUri,
```
hckenmiller: So based on this statement, we are using either implicit flow or authorization code flow, but the Client allowedGrantType is only Implicit.

Fabric.Identity.IntegrationTests/IntegrationTestsFixture.cs
```diff
@@ -175,21 +185,42 @@ protected static IdentityDbContext IdentityDbContext
             }
         }
 
-        public HttpClient HttpClient { get; }
+        private readonly TaskCompletionSource<HttpClient> _httpClientTaskCompletionSource;
+
+        public Task<HttpClient> HttpClient => _httpClientTaskCompletionSource.Task;
 
         private TestServer CreateIdentityTestServer(string storageProvider)
```
hckenmiller: With a couple test servers getting created, what if we consolidate the code into a method and create the TestServer with this method and only the custom settings for each test server should be different.

Fabric.Identity.IntegrationTests/ServiceTests/IdentityConfigurationProviderTests.cs
```diff
@@ -23,10 +25,14 @@ public void GetAppConfiguration_ReturnsAppConfig_Success()
...
             var decryptionService = new DecryptionService(mockCertificateService);
-            var identityConfigurationProvider = new IdentityConfigurationProvider();
+            var config = new ConfigurationBuilder()
+                .AddJsonFile(Path.Combine(directory.FullName, "appsettings.json"))
+                .Build();
+            var identityConfigurationProvider = new IdentityConfigurationProvider(config);
+            
 
             // Act
-            var appConfig = identityConfigurationProvider.GetAppConfiguration(directory.FullName, decryptionService);
```
hckenmiller: I noticed in the code that this string param was removed, is this not necessary in our setup to distinguish IdentityProviders?

Fabric.Identity.UnitTests/Validation/StubTokenValidator.cs
```diff
@@ -7,6 +8,7 @@ public class StubTokenValidator : ITokenValidator
     {
         public TokenValidationResult AccessTokenValidationResult { get; set; } = new TokenValidationResult();
         public TokenValidationResult IdentityTokenValidationResult { get; set; } = new TokenValidationResult();
+        public TokenValidationResult RefreshTokenValidationResult { get; set; } = new TokenValidationResult();
 
         public Task<TokenValidationResult> ValidateAccessTokenAsync(string token, string expectedScope = null)
```
hckenmiller: why are parameters being passed, but not used?

Fabric.Identity.UnitTests/Validation/StubTokenValidator.cs
```diff
@@ -17,6 +19,11 @@ public Task<TokenValidationResult> ValidateAccessTokenAsync(string token, string
             bool validateLifetime = true)
         {
             return Task.FromResult(IdentityTokenValidationResult);
-        }
+        }
+
+        public Task<TokenValidationResult> ValidateRefreshTokenAsync(string token, Client client = null)
```
hckenmiller: why are parameters being passed, but not used?

Fabric.Identity.API/Startup.cs
```diff
@@ -192,13 +213,12 @@ public void ConfigureServices(IServiceCollection services)
...
             IApplicationBuilder app,
             IHostingEnvironment env,
-            ILoggerFactory loggerFactory,
             IDbBootstrapper dbBootstrapper)
         {
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
-                _loggingLevelSwitch.MinimumLevel = LogEventLevel.Verbose;
+                LogFactory.LoggingLevelSwitch.MinimumLevel = LogEventLevel.Verbose;
```
hckenmiller: should this be something exposed in web.config or appsettings.json to allow the type of logging?

Fabric.Identity.API/Persistence/CouchDb/Stores/CouchDbResourceStore.cs
bskeen: I wasn't exactly sure if we were or not. I assume we aren't, but I was mostly focused on getting IdPSS moved over without much thought to removing things we don't need.

Fabric.Identity.API/Services/Azure/AzureActiveDirectoryClientCredentialsService.cs
bskeen: Kind of. IdPSS had 4 different search endpoints, and those are reflected in the different IdP services here. We can probably remove some of the old functionality from Identity after this release.

Fabric.Identity.API/scripts/BenTestLog.txt
bskeen: Yes, this should be removed. I will take care of that.

Fabric.Identity.API/web.config
bskeen: This should be fixed in master, but I will make sure of that before merging.

Fabric.Identity.API/Program.cs
bskeen: .Net Core 2.2 does things a little differently, and the upgrade guides say to do this. That being said, most of what is in the Main method is due to the Serilog upgrades.

Fabric.Identity.API/Services/Azure/AzureActiveDirectoryClientCredentialsService.cs
hckenmiller: Yah I can see the changes from a nancy controller to an mvc like controller.

PR #397, opened 2019-10-28
Clean up identity install artifacts

ctediz:
Clean up identity install artifact

Views folder is causing issues with identity after upgrading to .net core 2.2

### DIFF
--- a/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
+++ b/Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
@@ -469,7 +469,7 @@ function Unlock-ConfigurationSections(){
     $manager.CommitChanges()
 }
 
-function Remove-IdentityViewsFolder {
+function Remove-IdentityFolder {
     param (
         [string] $path
     )
@@ -479,7 +479,7 @@ function Remove-IdentityViewsFolder {
         # Clean subfolders
         Get-ChildItem $path -Recurse | Remove-Item -Recurse
         # Clean Folder
-        Remove-Item $path    
+        Remove-Item $path
     }
     elseif (Test-Path -Path $path -PathType Leaf) {
         Write-DosMessage -Level Information -Message "Deleting file: $path."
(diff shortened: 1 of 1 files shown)


Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -469,9 +469,29 @@ function Unlock-ConfigurationSections(){
...
+        Write-DosMessage -Level Information -Message "Deleting file: $path."
+        # Clean Item
+        Remove-Item $path
+    }
+}
+
 function Publish-Application([System.Object] $site, [string] $appName, [hashtable] $iisUser, [string] $zipPackage, [string] $assembly){
     $appDirectory = [io.path]::combine([System.Environment]::ExpandEnvironmentVariables($site.physicalPath), $appName)
     New-LogsDirectoryForApp $appDirectory $iisUser.UserName
+    Remove-IdentityViewsFolder -path "$appDirectory\Views"
```
hckenmiller: So views is a folder left over from .net core 1.1?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
```diff
@@ -469,9 +469,29 @@ function Unlock-ConfigurationSections(){
     $manager.CommitChanges()
 }
 
+function Remove-IdentityViewsFolder {
```
bskeen: I understand that you created this to remove the Identity Views folder, but the code inside the function seems more general purpose to me. Maybe this could be named `Remove-Folder` or something to that effect?

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: Yes it is

Fabric.Identity.API/scripts/Install-Identity-Utilities.psm1
ctediz: Slightly concerned about function name conflicts by going a route like that

hckenmiller: Good suggestion

ctediz: Could be something like Remove-IdentityFolder(?)

bskeen: I think that's a great idea.

hckenmiller: It removes files and folders, but we only care that the views folder is removed.

PR #399, opened 2019-10-31
Fix Window AD DI issues

ctediz:


Fabric.Identity.API/Startup.cs
```diff
@@ -115,8 +117,17 @@ public void ConfigureServices(IServiceCollection services)
                 .AddSingleton<PolicyProvider>()
                 .AddTransient<IHealthCheckerService, HealthCheckerService>()
                 .AddTransient<ICorsPolicyProvider, DefaultCorsPolicyProvider>()
+                .AddLocalization(opts => { opts.ResourcesPath = "Resources"; })
                 .AddFluentValidations();
 
+            var mapperConfig = new MapperConfiguration(cfg =>
```
hckenmiller: So was the current culture localization and auto mapper having issues as well, or did you notice they weren't getting registered and didn't want to have issues because of it.

ctediz: Ya the mapper issue showed up later in the process once the SearchServices were registered.  The localization issue showed up when trying out azure

PR #403, opened 2019-11-04
Add decryption of azure secrets

ctediz:
Add decryption of AzureActiveDirectoryClientSettings's secret

### DIFF
--- a/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
+++ b/Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
@@ -77,6 +77,17 @@ namespace Fabric.Identity.API.Configuration
                     appConfiguration.AzureActiveDirectorySettings.ClientSecret,
                     appConfiguration.SigningCertificateSettings);
             }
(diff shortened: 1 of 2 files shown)


Fabric.Identity.API/Configuration/IdentityConfigurationProvider.cs
```diff
@@ -77,6 +77,17 @@ private void DecryptEncryptedValues(IAppConfiguration appConfiguration, Decrypti
                     appConfiguration.AzureActiveDirectorySettings.ClientSecret,
                     appConfiguration.SigningCertificateSettings);
             }
+
+            if (appConfiguration.AzureActiveDirectoryClientSettings?.ClientAppSettings != null)
```
hckenmiller: So did we have this code before the .net core 2.2 upgrade but got missed with the changes?

ctediz: This, or the equivalent was in IdPSS, it was just missed when porting over

PR #408, opened 2019-11-19
Removed element <azureSecretName> which was causing an error in the scr…

tvidrine:
…ipt.  The file now matches the documentation.


tvidrine:
There were no errors on the previous install due to the test being done with a modified azureSettings.config file based on the documentation.  During this round of testing, I used the file included in the install which caused an error due to the azureSecretName not being found.

PR #413, opened 2019-11-21
User/tim/fix secret with invalid characters

tvidrine:


Fabric.Identity.API/Management/BaseController.cs
```diff
@@ -25,7 +26,15 @@ protected BaseController(AbstractValidator<T> validator, ILogger logger)
         }
 
         public Func<string> GeneratePassword { get; set; } =
-            () => Convert.ToBase64String(Guid.NewGuid().ToByteArray()).Substring(0, 16);
+            () =>
```
hckenmiller: It's probably ok, to have an illegal list, but I wonder if we could escape/remove any special characters so it is only letters and numbers.

tvidrine: I would be good with that as well to make sure we aren't missing anything that was tested.

